PR #47, opened 2017-02-18
Move all githubusercontent assets to S3

godfrzero:
Fixed fragility.


abhijitkane:
@godfrzero How does one upload into s3?

godfrzero:
One must go to S3 and upload directly. Not as "friendy" as before, but...well...yeah. 😕

shamasis:
Why not configure a webhook/Travis to upload to s3 using some form of encrypted access stored in Travis.

PR #52, opened 2017-02-23
Prefer the native app

loopDelicious:
Update copy to differentiate between native app and Chrome app
Increase prominence of native app
Update Postman Cloud copy to Postman Pro

@godfrzero would you please code review?


loopDelicious:
@godfrzero rebased, can you check it now?

PR #58, opened 2017-03-24
Update 2016-07-11-ignoring_ssl.md

czardoz:
This warning can be safely removed now.


godfrzero:
@czardoz Why is it okay to remove this now?

czardoz:
@godfrzero Because we no longer ignore SSL for all requests, just the one's that are sent by runtime.

PR #70, opened 2017-05-10
Updated sandbox docs with the version numbers of libraries relative to chrome vs native app

vegetableman:
cc @shamasis Please review the version numbers and the link to library set for `xml2json`.
cc @numaanashraf

### DIFF
--- a/_posts/06_api_testing_and_collection_runner/2016-07-11-sandbox.md
+++ b/_posts/06_api_testing_and_collection_runner/2016-07-11-sandbox.md
@@ -11,13 +11,13 @@ The Postman Sandbox is a JavaScript execution environment that is available to y
 
 ### Commonly used libraries and utilities
 
-* [Lodash][0]: JS utility library
-* [cheerio][1]: A fast, lean implementation of the core jQuery API (available in versions 4.6.0 and up)
+* [Lodash][0]: JS utility library  [ chrome `v3.6.0` | native `v4.17.2` ]
+* [cheerio][1]: A fast, lean implementation of the core jQuery API (available in versions 4.6.0 and up) [ native `v0.22.0` ]
 * [BackboneJS][2] **Deprecated**: Provides simple models, views, and collections. This will be removed in future versions of the sandbox.
-* [SugarJS][3]: Extends native JS objects with useful methods
-* [tv4 JSON schema validator][4]: Validates JSON objects against v4 of the json-schema draft
-* [CryptoJS][5]: standard and secure cryptographic algorithms. Supported algorithms: AES, DES, EvpKDF, HMAC-MD5, HMAC-SHA1/3/256/512, MD5, PBKDF2, Rabbit, SHA1/3/224/256/512, TripleDES
-* `xml2Json(xmlString)`: This function behaves the same in Newman and Postman
+* [SugarJS][3]: Extends native JS objects with useful methods [ `v1.4.1`]
+* [tv4 JSON schema validator][4]: Validates JSON objects against v4 of the json-schema draft [ `v1.2.7` ]
+* [CryptoJS][5]: standard and secure cryptographic algorithms. Supported algorithms: AES, DES, EvpKDF, HMAC-MD5, HMAC-SHA1/3/256/512, MD5, PBKDF2, Rabbit, SHA1/3/224/256/512, TripleDES [ chrome `v3.1.2` | native `v3.1.9-1` ]
+* [xml2Json][6]: `xml2Json(xmlString)` This function behaves the same in Newman and Postman [ `v0.4.17` ]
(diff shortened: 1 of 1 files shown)


vegetableman:
##Lodash Changelog

# 3.10.1 (native) - 3.10.6 (chrome)

```
v3.10.1

Aug. 4, 2015 — Diff — Docs

Ensured func is a function before calling getData
Ensured _.clone provides the correct number of arguments to customizer
Ensured _#reverse doesn’t modify the original array when called after _#slice
v3.10.0

June 30, 2015 — Diff — Docs

Added _.ceil, _.defaultsDeep, _.floor, _.modArgs, & _.round
Ensured _.bind works with all built-in constructors
Ensured _#concat treats sparse arrays as dense
Ensured _.chunk floors size values
Ensured _.debounce & _.throttle reset lastCall after cancelling
Ensure _.flow & _.flowRight work with functions combined with _.first
Ensured _.indexOf returns -1 for unmatched binary searches
Ensured _.noConflict operates on root & not context
Made _.escapeRegExp more robust
Made _.sortByOrder support orders of “asc” & “desc”
Optimized _.flatten, _.max, & _.min
Removed fallbacks for _.isArguments, _.isFunction, & _.isPlainObject
Simplified lazy evaluation support
Simplified _.isElement, _.isFinite, _.isNative, _.now, _.parseInt, & _.sum
v3.9.3

May 26, 2015 — Diff — Docs

Made parseFloat assigned from the context param of _.runInContext
Ensured _.set handles non-index property names that start with numbers correctly
v3.9.2

May 24, 2015 — Diff — Docs

Made isLaziable work with wrapped lodash methods
Optimized an early exit case in _.isEqual
Optimized _.sample
v3.9.0

May 19, 2015 — Diff — Docs

Added _.gt, _.gte, _.lt, & _.lte
Added support for an ImmutableMap to _.memoize.Cache
Avoided using require in source comments to improve browserify build times
Ensured baseCreate works in ExtendScript
Ensured customizer results are respected by _.isEqual
Ensured LodashWrapper.prototype.thru exists before creating a wrapper in flow
Ensured _.bind works with ES2015 class constructors
Ensured _.get can return null values
Ensured _.intersection works with a single array
Ensured _.has treats sparse arrays as dense
Ensured _.merge skips undefined array values if a destination value exists
Made null sort behind undefined & NaN
Made _.eq an alias of _.isEqual
Optimized object comparisons in _.isEqual
Optimized _.max & _.min when invoked with iteratees
Optimized _.pullAt & _.remove
Used hasOwnProperty for the creation of reIsNative to avoid issues with core-js
v3.8.0

May 1, 2015 — Diff — Docs

Added _.mapKeys, _.unzipWith, & _.zipWith
Made _.difference, _.intersection, _.uniq, & _.xor accept array-like values
Ensured empty brackets are ignored by _.get & _.set
Ensured baseAt, basePullAt, & pullAt handle nullish values
Ensured baseGet only returns undefined for incomplete paths
Ensured _.padLeft & _.padRight handle empty strings correctly
Made _.isEqual treat -0 & 0 as equivalent
Narrowed bitmask checks in _.flow & _.flowRight
Optimized lazy slice for start of 0
Optimized “flatten” methods
Restricted Object.assign use to strict mode only
v3.7.0

Apr. 15, 2015 — Diff — Docs

Added _.get, _.method, _.methodOf, & _.set
Avoided a JIT bug in Safari on at least iOS 8.1-8.3 ARM64
Ensured _.intersection of a single array returns an empty array
Ensured _.remove mutates array after the predicate pass
Ensured methods like _.has & _.get can access index values of strings
Made createAssigner & _.sortByAll use _.restParam
Made _.add coerce augend & addend params to numbers
Made _.assign use built-in Object.assign when available
Made _.inRange swap start & end params when start is greater than end
Added deep path support to methods like _.has, _.get, _.callback, _.invoke, _.matchesProperty, _.method, _.methodOf, _.property, _.propertyOf, _.result, & _.set
v3.6.0

Mar. 25, 2015 — Diff — Docs

Added _.restParam
Added isIterateeCall guards to _.every, _.includes, & _.some
Added support for iteratee & thisArg params to _.sum
Added support for shortcut fusion optimizations to _.flow & _.flowRight
Ensured lodash bundled by r.js runs in a web worker
Ensured _.deburr removes combining diacritical marks
Ensured _.difference is based on the values of the first param only
Expanded metadata optimizations to more combinations of curried functions
Made _.matches work with non-plain objects & match inherited properties
Optimized _.findLast, _.reduce, & _.reduceRight
Reduced code duplication with internal method creator functions
```

vegetableman:
CryptoJS is not maintaining change-logs from version `3.1.2` to `3.1.9-1`

vegetableman:
@loopDelicious Could you verify and merge this changes? Checkout the attached screenshot for the changes added:
screen shot 2017-05-15 at 12 43 00 pm

PR #96, opened 2017-06-21
Remove Sync - Getting started page

madebysid:
@loopDelicious


loopDelicious:
@madebysid any chance we can resolve this with a server-side redirect instead? unless we are confident that there are no active links to this page (from external sources), would prefer not to delete it.

fullpoint:
Good point -

madebysid:
@loopDelicious Can we redirect to `https://www.getpostman.com/docs/postman/launching_postman/syncing`?

loopDelicious:
@madebysid yes, that redirect works for me.

PR #111, opened 2017-07-25
Add search

loopDelicious:
@godfrzero / @abhijitkane please don't merge. can you please take a look at this? having trouble with the front end search display. algolia took a look and the code is working on their jsfiddle: https://jsfiddle.net/bobylito/0okx8ut7/. Do you know why this would NOT work with our config?


loopDelicious:
@godfrzero could you please review?

Add search to home page and sidebar.

_includes/_search.html
```diff
@@ -0,0 +1,101 @@
+<script>
+
+'use strict';
+/* global instantsearch */
+
+const search = instantsearch({
```
godfrzero: Not sure how stable `const` is in the wild. AFAIK, browser behavior will still be a bit flaky: some will treat it as a var, others will treat it as a constant but within a function scope.

If this ends up causing a problem at some point, it wouldn't hurt to just replace it with `var`

_includes/_search.html
```diff
@@ -0,0 +1,101 @@
...
+);
+
+const container = document.querySelector('#pagination');
+const paginationWidget = instantsearch.widgets.pagination({
+    container: '#pagination',
+    cssClasses: {
+        root: ['pagination', 'search-results'],
+        active: 'active'
+    },
+    hitsPerPage: 5
```
godfrzero: Why `5` here and `6` in the previous case?

_includes/_search.html
loopDelicious: Makes sense, replaced all.

_includes/_search.html
loopDelicious: `hitsPerPage` doesn't seem to do what I'd expect, will remove property in both instances.

PR #133, opened 2017-09-21
fix page 401 error when run locally with dev server.

66eli77:


godfrzero:
Only sure-fire way to make sure this works would be test it out on a fork. It shouldn't break anything according to the Jekyll documentation, but we can't take the risk 😛

PR #135, opened 2017-09-26
Improved Sandbox library reference

kunagpal:
* Removed duplicate `util` from native module list
* Clarified Lodash v4 usage.


kunagpal:
@loopDelicious

PR #163, opened 2017-11-02
Postman Accounts (switching user accounts)

btwisted:
Added switching user profiles info.
Links from existing documentation were broken. They started with /docs/etc. When I tried to use that format I repeated got a 404 errors message. Out of frustration I hard coded the previous and new links. Feel free to reconfigure the links as you want. At some point we can get together and you can show me what I'm doing wrong.


loopDelicious:
@btwisted can you update the pull request so that the base is `develop`? flip them around.

For the internal links (the ones that previously started with /docs/...), they will work when the site is built. They are relative links. 

The only ones that should be hard coded are the external links.

PR #183, opened 2017-11-22
Mid range updates to Navigating Postman

btwisted:
Needs style review, different animated images, and so on


btwisted:
Basically experimentation phase

PR #918, opened 2018-02-19
Update 2018-01-22-using_workspaces.md

btwisted:
In using workspaces, title of publish action should be "Publishing Collections" not workspaces.


btwisted:
resolved a conflict here! just want to note in case a problem arises.

PR #1039, opened 2018-03-09
Persist sidebar

loopDelicious:
- update nokogiri version for security vulnerability
- replace deprecated `algoliasearch-jekyll` with latest `jekyll-algolia`
- exclude v5 directory from indexing 
- persist sidebar accordion nav to highlight current, active page

@godfrzero could you please review? 🐮 

2 areas of interest:
- staging site doesn't have a server-side redirect from `/docs/` to `/docs/v6`, but I think docs homepage should work in production
- I excluded v5 from being indexed in the `_config.yml`, not currently updated on local, but I believe the site will reindex when we merge to master as travis script.


_includes/_search.html
```diff
@@ -3,14 +3,14 @@
 'use strict';
 /* global instantsearch */
 
-var search = instantsearch({
+const search = instantsearch({
```
godfrzero: I don't think we should be using const/let in our web code yet. IE 10 still has users and this won't function there. 😕

_includes/_search.html
```diff
@@ -48,17 +48,17 @@
 {% endraw %}
 
 {% raw %}
-var hitTemplate =
+const hitTemplate =
     '<div class="hit">' +
         '<div class="infos">' +
             '<div class="hit-title"><strong><a href={{site.pm.root}}/docs{{url}}>{{title}}</strong></a></div>' +
-            '<div class="hit-excerpt">{{{_highlightResult.text.value}}}</div>'+
+            '<div class="hit-excerpt">{{{_highlightResult.html.value}}}</div>'+
```
godfrzero: Why was this changed?

_includes/_search.html
loopDelicious: good call. updated.

_includes/_search.html
loopDelicious: after updating to the new search module `jekyll-algolia`, the previously used property was undefined.

godfrzero: 👍

PR #1116, opened 2018-03-20
Btw remove ent beta

btwisted:
Removed the "Apply for beta access" topic from Enterprise--the topic itself and the json.map entry.


fullpoint:
What happened with this?

PR #1348, opened 2018-07-16
Update to intro_monitors.md

chrisdeso:
Updated the total monitor run time from two (2) minutes to five (5).


chrisdeso:
@captchadex FYI☝️

PR #1423, opened 2018-09-25
Updated Settings Screen - gh1419

ArifPMan:
Please review and provide feedback..


ArifPMan:
In looks like this in prod:

image

PR #1455, opened 2018-10-22
Custom methods

ArifPMan:
Harry, please review the first draft. I have so many doubts still. Let's discuss..

### DIFF
--- a/v6/postman/customizing_postman.md
+++ b/v6/postman/customizing_postman.md
@@ -6,9 +6,9 @@ warning: false
 
 ### Creating Custom Methods
 
-Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custome Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. This feature lets you to save/delete the custom method and it also allows you to delete default methods thereby giving the entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there and type your custom method name and save it. It's that simple! The following screen illustrates this procedure. 
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custom Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. This feature lets you to save/delete the custom method besides allowing you to delete default methods thereby giving entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there, type your custom method name and save it. You can also send a request without saving this custom method. It's that simple! The following screen illustrates this procedure. 
 
-[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)
+[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods4.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods4.gif)
 
 **Note:** You can only save your custom methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. However, you are also not limited to use any request having a custom method that is not saved in the list. The only constraint is that it will not appear in the method drop-down.   
 
@@ -16,15 +16,13 @@ If you are in a team workspace and creating and using custom methods, the behavi
 
 1. When importing collections having requests with custom methods, you'll still be able to view these custom methods and be able to save by clicking the custom method.
 
-2. You must be on the same version of the app in order for you to effectively use this feature. 
+2. You must be on the same version of the app or above (6.5.0)in order for you to effectively use this feature. 
 
-3. If you import collections containing requests with custom methods in an older version of the Postman app, the custom methods can only be viewed and consumed but can't be saved. This behavior is influenced by the Sync feature.
+3. Likewise, even if you are in the older version of Postman app, you'll still be able to view and consume custom methods from the History tab if your team member had created custom methods using the newest version of the app. However, you'll not be able to save these custom methods.
 
-4. Likewise, even if you are in the older version of Postman app, you'll still be able to view and consume custom methods from the History tab if your team member had created custom methods using the newest version of the app. However, you'll not be able to save these custom methods.
+4. If you're in a team workspace and send a request by creating a custom method in it, your team members will not be able to view it unless they are on the same version of the app.
 
-5. If you're in a team workspace and send a request by creating a custom method in it, the history created cannot be viewed by your team members unless they are on the latest app.
-
-**Note:** Mock servers currently don't support custom methods. 
+**Note:** Creating a mock with a custom method is currently not supported by Postman.
 
 
 
--- a/v6/postman/customizing_postman.md
+++ b/v6/postman/customizing_postman.md
@@ -6,7 +6,7 @@ warning: false
 
 ### Creating Custom Methods
 
-Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custome Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. This feature also lets you to save/delete the custom method. This feature also allows you to delete default methods thereby giving the entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there and type your custom method name and save it. It's that simple! The following screen illustrates this procedure. 
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custome Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. This feature lets you to save/delete the custom method and it also allows you to delete default methods thereby giving the entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there and type your custom method name and save it. It's that simple! The following screen illustrates this procedure. 
 
 [![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)
 
--- a/v6/postman/customizing_postman.md
+++ b/v6/postman/customizing_postman.md
@@ -18,7 +18,7 @@ If you are in a team workspace and creating and using custom methods, the behavi
 
 1. When importing collections having requests with custom methods, you'll still be able to view these custom methods and be able to save by clicking the custom method.
 
-2. You must be on the same version of the app or above (6.5.0) in order for you to effectively use this feature. 
+2. You must be on the same version of the app (6.5.0) or above in order for you to effectively use this feature. 
 
 3. Likewise, even if you are in the older version of Postman app, you'll still be able to view and consume custom methods from the History tab if your team member had created custom methods using the newest version of the app. However, you'll not be able to save these custom methods.
 
--- a/v6/postman/customizing_postman.md
+++ b/v6/postman/customizing_postman.md
@@ -6,25 +6,25 @@ warning: false
 
 ### Customizing Request Methods
 
-You can customize request methods to meet corporate,  team-wide and other specific requirements as needed. Once you create your own request methods, you'll be able to send/save them. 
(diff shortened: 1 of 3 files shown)


v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods2.gif)
+
+**Note:** You can only save your custom methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. However, you are also not limited to use any request having a custom method that is not saved in the list. The only constraint is that it will not appear in the method drop-down.   
+
+If you are in a team workspace and creating and using custom methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When importing collections having requests with custom methods, you'll still be able to view these custom methods and be able to save by clicking the custom method.
+
+2. You must be on the same version of the app in order for you to effectively use this feature.
```
harryi3t: not necessarily same, it should be equal or above the version in which this feature will be released (6.5.0)

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Note:** You can only save your custom methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. However, you are also not limited to use any request having a custom method that is not saved in the list. The only constraint is that it will not appear in the method drop-down.   
+
+If you are in a team workspace and creating and using custom methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When importing collections having requests with custom methods, you'll still be able to view these custom methods and be able to save by clicking the custom method.
+
+2. You must be on the same version of the app in order for you to effectively use this feature. 
+
+3. If you import collections containing requests with custom methods in an older version of the Postman app, the custom methods can only be viewed and consumed but can't be saved. This behavior is influenced by the Sync feature.
```
harryi3t: NO! you can't import collections with custom methods in an older app. THose requests will not be imported

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
+---
+title: "Customizing Postman"
+page_id: "customizing_postman"
+warning: false
+---
+
+### Creating Custom Methods
+
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custome Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. This feature lets you to save/delete the custom method and it also allows you to delete default methods thereby giving the entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there and type your custom method name and save it. It's that simple! The following screen illustrates this procedure.
```
numaanashraf: typo `Custome`

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+If you are in a team workspace and creating and using custom methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When importing collections having requests with custom methods, you'll still be able to view these custom methods and be able to save by clicking the custom method.
+
+2. You must be on the same version of the app in order for you to effectively use this feature. 
+
+3. If you import collections containing requests with custom methods in an older version of the Postman app, the custom methods can only be viewed and consumed but can't be saved. This behavior is influenced by the Sync feature.
+
+4. Likewise, even if you are in the older version of Postman app, you'll still be able to view and consume custom methods from the History tab if your team member had created custom methods using the newest version of the app. However, you'll not be able to save these custom methods.
```
harryi3t: user will be able to see them but they can't use it
Can't save the request with custom method
They can open the history (with custom method) in a tab, when they click on send, the request will be sent but a history won't be created

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+5. If you're in a team workspace and send a request by creating a custom method in it, the history created cannot be viewed by your team members unless they are on the latest app.
```
harryi3t: same with the saved requests as well

v6/postman/customizing_postman.md
harryi3t: but in older app, history won't be created for custom methods (as mentioned in next point (5))

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Note:** Mock servers currently don't support custom methods.
```
harryi3t: Not sure about the terminology here. But, I think we just call them mock.
Something like
```
**Note:** Creating a mock with a custom method is not supported
```

ArifPMan:
Changes done. Numaan, please review and approve

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
+---
+title: "Customizing Postman"
+page_id: "customizing_postman"
+warning: false
+---
+
+### Creating Custom Methods
```
numaanashraf: IMO, this feature is not just about *creating* methods. Let's call this section, 'Customizing Request Methods'.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custom Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests.
```
numaanashraf: This para feels out of place in docs. More fit for the blog post. These seems to be a recurring concern - let me know if more clarification is needed.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+title: "Customizing Postman"
+page_id: "customizing_postman"
+warning: false
+---
+
+### Creating Custom Methods
+
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custom Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. 
+
+This feature lets you to save/delete the custom method besides allowing you to delete default methods thereby giving entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there, type your custom method name and save it. You can also send a request without saving this custom method. It's that simple! The following screen illustrates this procedure.
```
numaanashraf: paraphrase `entire user control to you`

numaanashraf: `custom method` is an internal jargon. let's see if we can rephrase this in the docs.

numaanashraf: "Working with this feature is a breeze" feels more like marketing than docs. Let's focus more on clearly explaining the feature. Sentences like these reads very weird when the user is trying to go through the docs and understand how to customize methods and work with this - which is the main purpose behind this doc.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+warning: false
+---
+
+### Creating Custom Methods
+
+Postman features can easily be extended and customized to meet corporate, team-wide and user-specific requirements. With a view to providing more user-control, Postman introduced **Custom Methods** feature. Using this feature, you can create your own custom methods. Once you create these methods, you'll be able to send/save your requests. 
+
+This feature lets you to save/delete the custom method besides allowing you to delete default methods thereby giving entire user control to you. Working with this feature is a breeze. Just point your cursor to the dropdown of the request method area, click there, type your custom method name and save it. You can also send a request without saving this custom method. It's that simple! The following screen illustrates this procedure. 
+
+[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods4.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods4.gif)
```
numaanashraf: let's show a screenshot with the app in a good state. in the current screenshot, the workspace mentions an unrelated feature name, lot of tabs open, response has a script error, etc.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+**Note:** You can only save your custom methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. However, you are also not limited to use any request having a custom method that is not saved in the list. The only constraint is that it will not appear in the method drop-down.
```
numaanashraf: paraphrase "However, you are also not limited to use any request having a custom method that is not saved in the list. The only constraint is that it will not appear in the method drop-down.   "

ArifPMan: custom method... you don't want this in the document... we can describe it in words... or can say you can create your own request methods

ArifPMan: Made all changes as requested. Let me know about custom method jargon. Only this is pending.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+If you are in a team workspace and creating your own methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When you import collections that have customized methods in them, you'll still be able to view them and be able to save by clicking the method.
+
+2. You must be on the same version of the app (6.5.0) or above in order for you to effectively use this feature. 
+
+3. Likewise, even if you are in the older version of Postman app, you'll still be able to view and consume the customized methods from the History tab if your team member created using the newest version of the app. However, you'll not be able to save these methods.
+
+4. If you're in a team workspace and send a request by creating your own method, your team members will not be able to view these methods unless they are on the same version of the app where these were created.
```
numaanashraf: "unless they are on the same version of the app where these were created." this is confusing. if teammates are on v6.5 and v6.6, this feature would still work without any issues. We should just mention that team members need to be on v6.5.0 and above.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,35 @@
...
+
+[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods5.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods5.gif)
+
+**Note:** You can save your customized methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. Postman also allows you to use requests containing your customized methods that are not saved in the list. However, these will not appear in the request method drop-down.   
+
+If you are in a team workspace and creating your own methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When you import collections that have customized methods in them, you'll still be able to view them and be able to save by clicking the method.
+
+2. You must be on the same version of the app (6.5.0) or above in order for you to effectively use this feature.
```
numaanashraf: they need not be on the "same version". they just need to be on version 6.5.0 and above.

v6/postman/customizing_postman.md
numaanashraf: I think we should get rid of no. 4. it contradicts point 3 - "even if you are in the older version of Postman app, you'll still be able to view". and the same version confusion detailed in previous comment.

v6/postman/customizing_postman.md
```diff
@@ -0,0 +1,33 @@
...
+
+[![custom methods1](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods5.gif)](https://s3.amazonaws.com/postman-static-getpostman-com/postman-docs/custom_methods5.gif)
+
+**Note:** You can save your customized methods locally, which means these are not synced to the Postman servers. This essentially means signing out will cause your data to be lost. Postman also allows you to use requests containing your customized methods that are not saved in the list. However, these will not appear in the request method drop-down.   
+
+If you are in a team workspace and creating your own methods, the behavior of this feature is a little dynamic so we recommend you to note the following points:
+
+1. When you import collections that have customized methods in them, you'll still be able to view them and be able to save by clicking the method.
+
+2. You must be on 6.5.0 version of the app or above in order for you to effectively use this feature.
```
harryi3t: @numaanashraf @ArifPMan I am trying to think of scenarios where even after reading these points users might have issues. 

Suppose user1 updates and start using this feature by creating a collection with some requests with customized methods. user2 does not see these requests (they get rejected by old the App), now user2 reads this doc and update to latest version. But, user2 still can't see those methods. **She has to sign-out and sign-in again to see those requests.** Should we mention this here, or is it too small of a detail to be here?

This wasn't an issue with other features where by just by updating your App you can start using and collaborating on the new features. Here, there might be an extra step involved which might be not very intuitive to the users.

numaanashraf: Let's not get too deep into transient issues users will face while a release is rolled out with the team. Its implied that new features on the latest version will have kinks if folks are working with new and old versions in parallel.

ArifPMan: Sure, I agree.

PR #1469, opened 2018-11-29
Removing {:target=“_blank”}

wsfuller:
With the new Learning Center we can no longer process {:target=“_blank”} like what was done with previous docs setup. However, any external link will have target and relationship attributes applied to it with a Markdown transforming plugin.

We are also fully migrating from getpostman.com/docs over to learning.getpostman.com over the next several days.


numaanashraf:
@wsfuller LGTM. Feel free to merge when you are ready.

wsfuller:
@numaanashraf Looks like I'm blocked from Merging

screen shot 2018-11-29 at 3 54 55 pm

numaanashraf:
@wsfuller could you try again? I've given you access.

PR #1535, opened 2019-03-11
Update and rename migrating_to_postman_v7.md to migrating_to_v7.md

abhijitkane:
The incorrect name/slug was causing problems updating the learning center


numaanashraf:
@ArifPMan Please take care of inconsistencies like these.

PR #1551, opened 2019-04-02
Updated poweredBy to false to remove Algolia logo

tristandenyer:
We pay for the Algolia search, so removing the logo.


loopDelicious:
🍰

PR #1574, opened 2019-04-26
setNextRequest scope: clearer guideline

p00j4:
**Change:** setNextRequest -> scope -> clearer message
**Why**
The current message says "postman.setNextRequest() is always executed at the end of the current script" which is ambiguous to understand as the current script can be "Pre-Request" or "Test" 
due to misunderstanding, I ended up investing time in debugging the actual cause. 

Discussion with Joyce [here for reference](https://community.getpostman.com/t/workflow-how-to-run-another-test-in-loop/2126/7?u=p00j).


ArifPMan:
We are not supporting the v5 branch. So, I'd like to delete this PR. Hope you're fine with this.

PR #1581, opened 2019-05-03
Fixed broken link for Markdown Styling collection in markdown documentation page

bhargavkaranam:


loopDelicious:
@bhargavkaranam looks like the link in v6 is different / broken

PR #1629, opened 2019-05-25
Updating Sync Screen & Content

ArifPMan:


tristandenyer:
@kaustavdm @DannyDainton c an you review for content/features to make sure they are correct? If so, please merge, verify in prod and delete branch.

PR #1714, opened 2019-07-25
[RUNTIME-937] Add settings docs for Max Response Size

arnitkun:
Added description for "max response size" option in the settings window.

### DIFF
--- a/v6/settings/settings.md
+++ b/v6/settings/settings.md
@@ -44,6 +44,11 @@ Prevent requests that return a 300-series response to be automatically redirecte
 * **Send anonymous usage data to Postman**
 Option to disable anonymous usage data, that we at Postman use to make Postman a better product. 
 
+* **Max response size**
+Option to limit the size of response (in Megabytes) that is received by postman from a server. Without setting a limit,
+postman app will give an error stating that the received response is large if it exceeds the default value of 50 MB.
+Note that setting this to a very large value to receive huge response might take postman a while.
+
 ### Theme
 
 Postman comes in two themes - light and dark. Choose the one that you're more comfortable working with.
--- a/v6/settings/settings.md
+++ b/v6/settings/settings.md
@@ -45,9 +45,9 @@ Prevent requests that return a 300-series response to be automatically redirecte
 Option to disable anonymous usage data, that we at Postman use to make Postman a better product. 
 
 * **Max response size**
-Option to limit the size of response (in Megabytes) that is received by postman from a server. Without setting a limit,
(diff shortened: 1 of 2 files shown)


v6/settings/settings.md
```diff
@@ -44,6 +44,11 @@ Prevent requests that return a 300-series response to be automatically redirecte
 * **Send anonymous usage data to Postman**
 Option to disable anonymous usage data, that we at Postman use to make Postman a better product. 
 
+* **Max response size**
+Option to limit the size of response (in Megabytes) that is received by postman from a server. Without setting a limit,
```
kaustavdm: Rephrase this as the limit of response size that Postman will attempt to render. Setting this to a higher value will make Postman use more memory to render.

v6/settings/settings.md
```diff
@@ -44,6 +44,11 @@ Prevent requests that return a 300-series response to be automatically redirecte
...
+postman app will give an error stating that the received response is large if it exceeds the default value of 50 MB.
```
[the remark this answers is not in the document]
kaustavdm: Postman does not really give an error, but tells you that the response is too large to handle. User can then choose to download the response.

v6/settings/settings.md
```diff
@@ -44,6 +44,11 @@ Prevent requests that return a 300-series response to be automatically redirecte
 * **Send anonymous usage data to Postman**
 Option to disable anonymous usage data, that we at Postman use to make Postman a better product. 
 
+* **Max response size**
+Option to limit the size of response (in Megabytes) that is rendered by Postman. If the limit is exceeded,
+Postman will inform the user that the received response is large (default limit: 50 MB) and provide options to increase the size limit or download the response.
+Note that setting this to a very large value to receive huge response will require greater amount of memory to render and may take some time.
```
saswatds: @kaustavdm  Technically, setting this to very large values (>200mb) impacts the performance of the app. The response might be very slow to render and in worst-case may even crash the app.
I am not very certain if we should put this point strongly or not. Your call. Other wise LGTM

kaustavdm: @arnitkun Rephrase the last sentence as "setting this to a high value can impact app performance".

kaustavdm:
@tristandenyer Can you confirm if this change is made to the right file? I see 2 different files in `./v6/` for settings with identical content. Holding off merging till you confirm.

tristandenyer:
@kaustavdm I don't see these changes in Production. I also don't see any updates having been logged in Contentful. Something isn't right. 🤔

kaustavdm:
@tristandenyer any pointers on where I can check for this? Is there any way to trigger the changes?

PR #1811, opened 2019-09-27
replaced all links to old s3 bucket

Ryan-McBride:
replaced all old image links to point to new assets endpoint.


ch264:
11 files were stuck in production. They were updated and republished.

PR #1848, opened 2019-10-15
Update navigating_postman.md to remove extra word

claireriley:
Removed "also"


claireriley:
Ignore this — leaving as a comment in a review instead.

Issue #1857, opened 2019-10-22
Visualizer doc extensions


SueSmith:
@kaustavdm thought we could talk through the visualizer docs here..

So have created the [feature/doc-style-guidelines](https://github.com/postmanlabs/postman-docs/tree/feature/doc-style-guidelines) branch which has the draft style guide and initial visualizer doc edits in it. 

The style guide is _definitely_ only a first draft and will likely need work.. in particular the [doc sections part](https://github.com/postmanlabs/postman-docs/blob/feature/doc-style-guidelines/DOCS_STYLE_GUIDE.md#document-sections) has been drafted with the visualizer doc in mind but is rough and will need extended later for different topics..

You'll see I've added / moved a few headings in the visualizer doc to get us started, but in general this is what I'd recommend:

* I'd lose the use cases heading but maybe try to incorporate the core info in there in the intro and maybe other sections, refocusing the language to be more about what you can do and how to do it and less on persuading people of its value if that makes sense
* As you mentioned, the main section walking through how to use visualizers could be extended as more of a practical walkthrough, with subheadings for steps etc
* As you'll see I moved two of the sections up into the tutorial-type bit (using libraries and accessing template data)
* Your extra stuff on templates etc could maybe extend the "try it out" section..?
* API and debugging sections look good to me but any extensions you had planned there are obviously handy (might be getting to the point where we need to think about additional pages if it starts getting too long but maybe that's for later)

Anyway hope all that makes sense and is helpful for now.. 😬

PR #1903, opened 2019-10-30
[DEVREL-55] Visualizer docs changes for GTM plan

kaustavdm:
@SueSmith I took the changes you mentioned in #1857 and applied them on the new LC structure, along with my changes.

You can see the changes [here](https://github.com/postmanlabs/postman-docs/blob/feature/devrel-55-visualizer-docs-redo/src/pages/docs/postman/sending-api-requests/visualizer.md).


SueSmith:
Just a note that we have these new images on s3 so can switch to that if it's a problem to pull them from static assets.

Issue #1856, opened 2019-10-21
`pm.collectionVariables` is not documented


jrop:
Ideally, `pm.collectionVariables` should be documented on this page:

https://learning.getpostman.com/docs/postman/scripts/postman_sandbox_api_reference/

arnitkun:
Thank you for letting us know. Yes, collectionVariables will be documented soon. Will update once it is done.

arnitkun:
The changes have been merged. Closing this issue.

PR #1921, opened 2019-11-04
Documentation for pm.collectionVariables

arnitkun:
This PR adds documentation around pm.collectionVariables, addressing #1856.

### DIFF
--- a/src/pages/docs/postman/environments-and-globals/variables-complete-reference.md
+++ b/src/pages/docs/postman/environments-and-globals/variables-complete-reference.md
@@ -73,7 +73,7 @@ pm.variables.get();
 
 ```
 
-Let’s consider a case in point. A collection has 500 requests in it and all of them have a URL. Let’s say if the URL gets changed then updating this change 500 times for 500 requests is near impossible. So, it’s a good idea to create a collection variable and use accordingly. Another recommendation here is if you have non-sensitive data which is repetitive, it is recommended to go for a collection variable. Remember one of the primary use cases of collection variables is that you can access them from a request within a collection.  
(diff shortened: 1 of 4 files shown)


kaustavdm:
@SueSmith Can you do an editorial pass on this PR?

src/pages/docs/postman/environments-and-globals/variables-complete-reference.md
```diff
@@ -73,7 +73,7 @@ pm.variables.get();
 
 ```
 
-Let’s consider a case in point. A collection has 500 requests in it and all of them have a URL. Let’s say if the URL gets changed then updating this change 500 times for 500 requests is near impossible. So, it’s a good idea to create a collection variable and use accordingly. Another recommendation here is if you have non-sensitive data which is repetitive, it is recommended to go for a collection variable. One point to note though is that you can’t set collection variables using scripts. Remember one of the primary use cases of collection variables is that you can access them from a request within a collection.  
+Let’s consider a case in point. A collection has 500 requests in it and all of them have a URL. Let’s say if the URL gets changed then updating this change 500 times for 500 requests is near impossible. So, it’s a good idea to create a collection variable and use accordingly. Another recommendation here is if you have non-sensitive data which is repetitive, it is recommended to go for a collection variable. Remember one of the primary use cases of collection variables is that you can access them from a request within a collection.
```
SueSmith: So am trying to remove first person from the docs ("Let's" in this case) - maybe rephrase something like:

> Collection variables can help manage large collections, for example if you need to change the URL for a collection with hundreds of requests in it—using a collection variable provides a single point of change. Collection variables help with any non-sensitive data that you need to repeat in multiple places. You can access collection variables from any request inside a collection.

arnitkun:
Changes made. Apparently a lot of things are written in the first person; this existed in the docs before the pull request.

PR #1955, opened 2019-11-08
[DEVREL-71] Document remaining mock server creation flows

### LINKED ISSUE #1498: Clarify how to use an environment with a mock server
Feedback from user: 

> When creating a Mock server, there is an option to select an existing Environment. However I can't figure out the purpose of this and can't see any effects either. A new environment is always created for the Mock. 

You can use an environment when you create a mock if you plan to use any environment variables in your mock (response, headers, url, etc). If not, then there's no need to use an environment. The default flow when you create a mock from the New+ button is to also create an environment - it's not really necessary, but you can see that an environment variable called `url` is used.

kaustavdm:
Documented remaining mock server flows through collection sidebar and history request.

**[Rendered](https://github.com/postmanlabs/postman-docs/blob/feature/devrel-71-mock-creation-gaps/src/pages/docs/postman/mock-servers/setting-up-mock.md#collection-sidebar)**

Fixes #1498

### DIFF
--- a/src/pages/docs/postman/mock-servers/setting-up-mock.md
+++ b/src/pages/docs/postman/mock-servers/setting-up-mock.md
@@ -35,11 +35,25 @@ contextual_links:
 warning: false
 ---
 
-[Mock Servers](/docs/postman/mock-servers/intro-to-mock-servers) in Postman let you simulate APIs. You can create mock servers from the Postman app, from the web dasboard and using the Postman API. You will need a Postman account to set up a mock server.
+[Mock Servers](/docs/postman/mock-servers/intro-to-mock-servers) in Postman let you simulate APIs. You can create mock servers from the Postman app, from the web dasboard, and using the Postman API. You will need a Postman account to set up a mock server.
 
 Mocks in Postman are tied to a collection. Postman matches requests and generates responses for mocks from the Examples in the requests of a collection. You can create a mock server even if you don't have an existing collection.
 
-## In-app
+## Contents
+
+* [Creating mock servers in-app](#creating-mock-servers-in-app)
+    * [Mocking from the New button](#mocking-from-the-new-button)
+    * [Mocking from the launch screen](#mocking-from-the-launch-screen)
+    * [Mocking from the Collections sidebar](#mocking-from-the-collections-sidebar)
+    * [Mocking a request from History](#mocking-a-request-from-history)
+    * [Mocking from Collections in Browse view](#mocking-from-collections-in-browse-view)
+* [Creating mock servers in the web dashboard](#creating-mock-servers-in-the-web-dashboard)
+* [Using HTTP access control for a mock](#using-http-access-control-for-a-mock)
+* [Using free mock server calls](#using-free-mock-server-calls)
+* [Editing mock servers](#editing-mock-servers)
+* [Next steps](#next-steps)
+
+## Creating mock servers in-app
 
 You can create a mock in several ways in the Postman app:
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
 
 **Note**: At the bottom, you can select "Show this window at launch" to indicate whether you want the Create New tab to display each time you open Postman.
 
+### Collection sidebar
+
+You can create a mock server from an existing collection in the left sidebar. To do that, expand the collection details pane, switch over to the "Mocks" tab in the pane, and click on the "Create a mock server" button.
```
SueSmith: So in general instead of putting UI features we want the user to interact with in quotes I'd switch to putting them in bold, so "the __Mocks__ tab" etc. 

Also I'd go for just "click" instead of "click on" - can shorten even further by removing "the ... button" at the end of this sentence although that's more discretionary, kind of depends on how intuitive the UI feature is.. That would be "click __Create a mock server__."

src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
...
+
+![Create new mock server from collection sidebar](https://assets.postman.com/postman-docs/mock-cnx-collection-sidebar.png)
+
+This will launch the "Create mock server" modal. Follow the same steps as earlier to complete the process.
```
SueSmith: Could we maybe make "the same steps as earlier" a bit more explicit in case the person has scanned to this point in the doc, e.g. "complete the mock server setup steps" with that text linking to the part up the page where those steps are..?

src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
...
+
+If you already have mock servers created from the same collection, you will see a button called "Add mock", instead of the "Create a mock server" button.
```
SueSmith: Same as above on the bold instead of quotes for "add mock" - I won't comment on all of these, if we can swap them all over to bold that would be great. For here maybe something like "you will see an __Add mock__ button instead of __Create a mock server__."

src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
...
+
+![Create new mock server from collection sidebar when there are existing mock servers](https://assets.postman.com/postman-docs/mock-cnx-collection-sidebar-existing.png)
+
+### History request
+
+You can mock a single request from the History tab in the left sidebar. To do that, switch to the "History" tab in the left sidebar, hover on an entry, and click on the button with three dots to expand the dropdown menu. Choose "Mock Request" from the dropdown.
```
SueSmith: Again for "click" instead of "click on"..

src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
...
+
+![Create new mock server from history request](https://assets.postman.com/postman-docs/mock-cnx-history-request.png)
+
+This will open up a modal with the screen titled "2. Set up the mock server". Pass the values similar to the previous steps. Click "Create Mock Server".
```
SueSmith: Maybe condense first sentence here to something like "Complete the details in the mock server setup modal" again linking to the steps outlined earier in the doc..?

src/pages/docs/postman/mock-servers/setting-up-mock.md
```diff
@@ -95,6 +95,38 @@ Follow the steps in the previous section.
...
+
+This process of creating a mock server will also create a collection with the same title as the mock server. This collection will have a copy of the same request that you chose from the History sidebar. The mock server created is tied to this new collection.
+
+![New collection created when creating mock server from history request](https://assets.postman.com/postman-docs/mock-cnx-history-request-created.png)
+
+### Collections in Browse view
+
+You can create a mock from an existing collection from the "Browse" view as well. Switch over to the "Browse" view by clicking the toggle button on the bottom right hand side of the Postman app. Change to the "Collections" tab for the current workspace. Click the button with the three dots called "View more actions" for the corresponding collection. Select "Mock Collection" from the dropdown.
```
SueSmith: This is definitely another discretionary thing but sometimes I avoid things like "as well" in case the person has scanned straight to this part of the doc and doesn't have context, but need to use your judgement case by case for whether it works..

SueSmith: I've been wondering how best to refer to the three dots buttons.. not sure but will share if I come up with anything useful.

SueSmith:
Wondering if this maybe fixes #1498 @kaustavdm ?

SueSmith:
Commenting to keep this open

src/pages/docs/postman/mock-servers/setting-up-mock.md
kaustavdm: The flow opens at different points depending on which flow the user chose. I will leave this on as is for now.

src/pages/docs/postman/mock-servers/setting-up-mock.md
kaustavdm: Same here. This needs the product glossary

kaustavdm:
@SueSmith I have made the changes. Sorry about the delay. Can you see if this is good to go? Also, yes, this should close #1498.

PR #1956, opened 2019-11-10
1452 added JWT term to copy

tristandenyer:
fixes 1452

Added JWT copy to page to come up in search results.


SueSmith:
I'm actually completely rewriting this page just now so may be easier to just add this into my redraft which is almost complete.

SueSmith:
Just to confirm I've added JWT to the bearer token section in the new version of this doc.

PR #1958, opened 2019-11-10
1652-updated command lines and links

tristandenyer:
fixes 1652

— updated command lines according to updated Docker page

### DIFF
--- a/src/pages/docs/postman/collection-runs/newman-with-docker.md
+++ b/src/pages/docs/postman/collection-runs/newman-with-docker.md
@@ -30,16 +30,16 @@ tags:
 
 ---
 
-This topic describes how to use Newman with Docker in these platforms:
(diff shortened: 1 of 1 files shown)


src/pages/docs/postman/collection-runs/newman-with-docker.md
```diff
@@ -52,16 +52,16 @@ $ docker run hello-world
 3\. Pull the Newman docker image.
 
 ```bash
-$ docker pull postman/newman:ubuntu
+$ docker pull postman/newman;
```
SueSmith: does this need the semicolon on the end of it?

tristandenyer: That's how it is presented here: https://hub.docker.com/r/postman/newman

SueSmith: cool I don't use docker but any other examples I could see didn't have a semicolon on them

PR #1962, opened 2019-11-11
Redraft of auth doc

### LINKED ISSUE #1679: Add callback URL to OAuth 2 instructions on Learning Center
Doc: https://learning.getpostman.com/docs/postman/sending_api_requests/authorization/#oauth-20

Leaving the callback URL empty may not be appropriate for all use-cases. The following URL can be used where required: `https://www.getpostman.com/oauth2/callback`

Would be good to enumerate some examples of where this may be required (e.g. Azure Directory Services).

SueSmith:
General redraft, also replaced images and added detail on auth flows etc. It's a really long doc so might be worth splitting up at some point..

Fixes #1679


SueSmith:
Will add detail to fix #1679

PR #1979, opened 2019-11-13
added Peter Evans link checker to pages folder

ch264:
Github Action Plugin: Peter Evans Link checker, checks for broken links and images in pages folder.


ch264:
This checker runs for too long. Will so more research on it.

PR #2041, opened 2019-12-11
Fix example for environment variable

Clem69B:
Fix example for environment variable usage in script : "variable_value" and not "variable_key" as second argument.


SueSmith:
@Clem69B Thank you for fixing this! :tada:

SueSmith:
Change should be live! Thanks again @Clem69B 🏆 🎊 🚀

PR #2042, opened 2019-12-11
Removed a localhost:8080 URL

jchekanoff:
A bad `localhost` URL was left from local development; removed it in favor of a directory link.


SueSmith:
@jchekanoff Thanks so much for fixing this! :confetti_ball:

SueSmith:
Your change should now be live! Thanks again @jchekanoff 🚀 🎉 🏆

PR #2053, opened 2019-12-17
Minor spelling fix

karlfus:


SueSmith:
@karlfus thanks so much! 🎉

SueSmith:
@karlfus this has been pushed to the live site so should be up soon, thanks again 🏆

PR #2068, opened 2019-12-20
added code_keyword to frontmatter

ch264:
added a code_keyword to frontmatter in docs yml file. That way we can capture terms from within <code> and <pre> tags that we want to have indexed by Algolia

### DIFF
--- a/src/pages/docs/postman/variables-and-environments/variables.md
+++ b/src/pages/docs/postman/variables-and-environments/variables.md
@@ -2,6 +2,7 @@
 title: "Using variables"
 order: 48
 page_id: "variables"
+code_keyword: "collectionVariables, iterationData, collectionVariables.set, collectionVariables.get" 
 contextual_links:
   - type: section
     name: "Prerequisites"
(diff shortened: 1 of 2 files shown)


src/pages/docs/postman/variables-and-environments/variables.md
```diff
@@ -2,6 +2,7 @@
 title: "Using variables"
 order: 48
 page_id: "variables"
+code_keyword: "collectionVariables, iterationData, collectionVariables.set, collectionVariables.get"
```
tristandenyer: I would like to suggest something less restrictive. This term(s) is meant to augment the page so that it shows up in search results for a better user experience. While I feel it is likely in the vast majority of the use case to be a term inside a codeblock, that may not always be the case.

I'd like to propose using `search_keyword` or `index_keyword` so as to allow for terms that are not just inside code blocks. What do you think?

ch264: I like search_keyword. @SueSmith are you ok with changing it to that?

SueSmith: oops sorry for some reason I thought I had responded to this already - yes works for me!

ch264: @SueSmith no worries. I will change it now :)

PR #2087, opened 2020-01-08
Feature/console improvements

loopDelicious:
Still need to confirm limit of request/response body - currently 300KB in docs @ritwijsinha 
And that youtube video - currently unlisted is ready to be listed @kaustavdm https://www.youtube.com/watch?v=XCfSYT9g0BU

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,8 @@ yarn-error.log
 # Jekyll
 .jekyll-metadata
 _site/
+
+package-lock.json
+package.json
+yarn.lock
+.idea/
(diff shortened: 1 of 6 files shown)


SueSmith:
@loopDelicious few wee things to mention:

* not sure if package-lock and package json files should be part of the commit..?
* as per style guide we've been trying to remove first person (just from docs, not blog posts or whatever), so "we" and "let's" would go but "you" is still ok
* references to ui elements the reader is clicking in bold, e.g. __Response Headers__

Hope that makes sense.. 😬

loopDelicious:
>- not sure if package-lock and package json files should be part of the commit..?
> - as per style guide we've been trying to remove first person (just from docs, not blog posts or whatever), so "we" and "let's" would go but "you" is still ok
> - references to ui elements the reader is clicking in bold, e.g. Response Headers

@SueSmith updated! for the last 2 bullets, i only updated the 2 primary debugging pages that I was already editing (not the remainder of the sections).

SueSmith:
@loopDelicious magic thanks! oh god yeah I wouldn't open the pandora's box of editing other bits lol 🙈

ch264:
actually all three should be removed from .gitignore: 
package-lock.json
package.json
yarn.lock

PR #2096, opened 2020-01-10
Revert "Revert "documentation for API relation validation""

SueSmith:
**Please don't merge this until the schema validation update is in prod**

### DIFF
--- a/src/pages/docs/postman/design-and-develop-apis/validating-elements-against-schema.md
+++ b/src/pages/docs/postman/design-and-develop-apis/validating-elements-against-schema.md
@@ -1,0 +1,131 @@
+---
+title: 'Validating Elements Against Schema'
+order: 91.1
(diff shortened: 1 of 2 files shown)


BrandonGordon126:
@arlemi One not-so-small nomenclature note: According to the [lexicon](https://postmanlabs.atlassian.net/wiki/spaces/PRODUCT/pages/1233552047/RFC-77+Postman+Lexicon), we should be using the term "elements" instead of "relations." 

cc: @SueSmith

arlemi:
@BrandonGordon126 Updated content and filename to match the lexicon.

@SueSmith Only changes made were replacing `relation` occurrences with `element`, it should still be good to go.

arlemi:
@SueSmith This is out in prod, ready to merge when you're ready!

PR #2097, opened 2020-01-10
Documentation for request validation

arlemi:
This is a first go at the request validation documentation. Any input highly appreciated :D

### DIFF
--- a/src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
+++ b/src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
@@ -17,7 +17,7 @@ This section outlines the following topics:
 
 ## Validating requests
 
-To execute validation on your requests, you need to link their collection to an API.
+If you want your requests to be validated, your collection must be linked to an API.
 
 A collection is linked to an API if you [generate it from a schema](docs/postman/design-and-develop-apis/the-api-workflow/#generating-a-collection-from-a-schema) or [add it as a relation to an existing API](/docs/postman/design-and-develop-apis/the-api-workflow/#developing-an-api).
 
@@ -62,7 +62,7 @@ This section lists each type of issue that validation may raise along with examp
 
 `Invalid type` issues occur when the basic type doesn't match, or when the basic type matches but the schema is wrong.
 
-> The path variable "petId" needs to be of type boolean, but we found `<integer>`
+> The path variable "petId" needs to be of type integer, but we found `Cooper`
 
 ### Missing in schema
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
```diff
@@ -0,0 +1,91 @@
...
+
+This section outlines the following topics:
+
+* [Validating requests](#validating-requests)
+* [Accessing issues](#accessing-issues)
+* [List of possible issues](#list-of-possible-issues)
+
+## Validating requests
+
+To execute validation on your requests, you need to link their collection to an API.
```
vvs11: This needs to be changed to something like:
If you want your requests to be validated, your collection must be linked to an API

src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
```diff
@@ -0,0 +1,91 @@
...
+* [Missing in schema](#missing-in-schema)
+* [Missing in request](#missing-in-request)
+* [Invalid body](#invalid-body)
+* [Body schema not found](#body-schema-not-found)
+
+### Invalid type
+
+`Invalid type` issues occur when the basic type doesn't match, or when the basic type matches but the schema is wrong.
+
+> The path variable "petId" needs to be of type boolean, but we found `<integer>`
```
vvs11: Can we change <integer> to something else? Here <integer> is shown because it is the type of the field as mentioned in the schema and the validation against it is failing because it is being read as a string. This needs to be fixed by the converter module and the actual value provided in the example should be displayed. Users usually don't use <integer> as a query params and it doesn't look good in the documentation

src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
```diff
@@ -0,0 +1,91 @@
...
+
+> The required path variable "name" was not found in the transaction
+
+### Invalid body
+
+`Invalid body` issues occur when the request body didn't match the one specified in the schema.
+
+### Body schema not found
+
+`Body schema not found` issues occur when no `application/json` schema was found for the response.
```
vvs11: should it be" `Body schema not found` issues occur when no `application/json` schema **is** found for the response."?
You might want to check this once.

src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Next steps
+
+Your issue summary includes the details to address any problems validating against your schema. You can edit the relevant components of your request and validate again to see if the issues have been resolved.
```
vvs11: can this be re-worded too? not very clear what is being conveyed

src/pages/docs/postman/sending-api-requests/validating-requests-against-schema.md
arlemi: Are you talking about the example, if so would something like this be better? 
> The path variable "petId" needs to be of type integer, but we found `Cooper`

arlemi:
@SueSmith Merging since this is now out in prod :D

PR #2099, opened 2020-01-10
included embedded approved video

tobiaswettstein:
Hello all
I have embedded the new approved video (with help from Christina). The video should be embedded instead of just a link so it is easier to be found and noticed.
Thanks all!
Tobi

### DIFF
--- a/src/pages/docs/postman/sending-api-requests/visualizer.md
+++ b/src/pages/docs/postman/sending-api-requests/visualizer.md
@@ -165,6 +165,15 @@ You can debug a visualization in Postman by right-clicking in the __Visualize__ 
 
 ## Next steps
 
-The [visualizer demo video](https://www.youtube.com/watch?v=Qj7j3QsY2ok) walks through the process of building visualizations in Postman.
+The visualizer demo video walks through the process of building visualizations in Postman.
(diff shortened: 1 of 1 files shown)


ch264:
@SueSmith @nicktran We can embed the video in the docs with an image of the video. Clicking on the image of the video will open the video in Youtube.

SueSmith:
Hm not sure it's obvious enough visually that this is a video, can we just embed it using html instead so it plays in the page? Pretty sure the docs used to have videos embedded directly..

SueSmith:
For reference this is what it looks like when I run locally:
Screenshot 2020-01-13 at 10 25 00

I would probably put it in a different part of the page too but can sort that once we figure out how it's getting embedded.

ch264:
@SueSmith I will look into how we can embed the video into html. I am not sure if that is at all possible with Markdown. I will get back with more info

SueSmith:
@ch264 I just meant drop it into the markdown page in html directly using the youtube embed code or whatever..

ch264:
@SueSmith, I created a new PR for the issue # 2098

https://github.com/postmanlabs/postman-docs/pull/2103

ch264:
@tobiaswettstein Thanks very much for the PR. We have found a way to embed videos in Markdown with Gatsby plugins. I will therefore close this PR.

PR #2104, opened 2020-01-14
Activity feed docs

arlemi:
First try at the activity feed docs. 

@sankalp0o, I ended up not adding the exhaustive list of the activities that are shown in the activity feed but if needed can add them afterwards.


arlemi:
Ignore this one 🤦‍♂

PR #2125, opened 2020-01-24
added API entity in activity feed filtering

arlemi:
This adds the notion of being able to filter by API entity in the activity feed.

ETA for prod is 27th of Jan.


arlemi:
I may have forgotten about this one.

PR #2127, opened 2020-01-24
Inline edit updates

SueSmith:
Updates to API documentation docs for inline edit.


SueSmith:
@Raja-Simha added a bit more detail after seeing this at demo day..

SueSmith:
@claireriley would you mind giving this a quick once over too? Is in prod so will just get it published then update later if any changes are required from squad..

PR #2132, opened 2020-01-27
added css to wrap <pre> tags content

ch264:
In order to fix the layout issues, the content within <pre> text is now wrapped.


SueSmith:
@ch264 k have changed the formatting on that one page, I don't think we should be dumping terminal output in the page content like that anyway so if it works from a layout perspective we should be fine..

ch264:
@SueSmith awesome, thanks so much. Merging now.

PR #2158, opened 2020-02-04
updating footer

ch264:
updating footer to look more like website footer.


src/components/Footer/Footer.data.json
```diff
@@ -64,7 +66,7 @@
   "copyright": [
     {
       "type": "notice",
-      "text": "© Postman, Inc. All rights reserved"
+      "text": "©2020 Postman, Inc. All rights reserved"
```
Ryan-McBride: Should we be hardcoding 2020 or should we insert the current year?

ch264: This will come from the API in future so will leave as is for now

PR #2175, opened 2020-02-07
Update variables.md

jyoung9:
I've added an example of accessing a variable from the body of a request. This is something I felt the documentation needed, as I have seen my initial confusion expressed by others on various discussion boards.

### DIFF
--- a/src/pages/docs/postman/variables-and-environments/variables.md
+++ b/src/pages/docs/postman/variables-and-environments/variables.md
@@ -258,7 +258,7 @@ Postman will send whatever value you currently have stored for the `cust_id` var
 http://pricey-trilby.glitch.me/customer?id=3
 ```
 
-Alternatively, a request body that accesses a variable by wrapping its reference in double-qoutes:
+Alternatively, you cou;d have a request body that accesses a variable by wrapping its reference in double-qoutes:
 
 ```
 { "customer_id" : "{{cust_id}}" }
--- a/src/pages/docs/postman/variables-and-environments/variables.md
+++ b/src/pages/docs/postman/variables-and-environments/variables.md
@@ -252,16 +252,16 @@ For example, you could have a request URL referencing a variable as follows:
 http://pricey-trilby.glitch.me/customer?id={{cust_id}}
 ```
 
-Or, a request body that accesses a variable by wrapping its reference in double-qoutes:
-
-```
-{ "customer_id" : "{{cust_id}}" }
-```
-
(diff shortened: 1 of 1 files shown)


src/pages/docs/postman/variables-and-environments/variables.md
```diff
@@ -252,6 +252,12 @@ For example, you could have a request URL referencing a variable as follows:
 http://pricey-trilby.glitch.me/customer?id={{cust_id}}
 ```
 
+Or, a request body that accesses a variable by wrapping its reference in double-qoutes:
```
SueSmith: Thanks so much for adding this! Think you could change the sentence to begin "Alternatively, you could have a request body..." just so that it doesn't start with the conjunction "Or"?

Could you also move your new section a little further down, to before the paragraph that starts "You can use variables in request URLs.." so that it doesn't break this section if that makes sense (since the next paragraph refers to the query parameter example)?

jyoung9:
Yes, of course. I'm happy to see my addition approved and so it's no
problem to make it fit in a little better. Thanks!

src/pages/docs/postman/variables-and-environments/variables.md
```diff
@@ -252,16 +252,16 @@ For example, you could have a request URL referencing a variable as follows:
...
-Or, a request body that accesses a variable by wrapping its reference in double-qoutes:
+Postman will send whatever value you currently have stored for the `cust_id` variable when the request runs. If `cust_id` is currently `3`, the request will be sent to the following URL including query parameter:
 
 ```
-{ "customer_id" : "{{cust_id}}" }
+http://pricey-trilby.glitch.me/customer?id=3
 ```
 
-Postman will send whatever value you currently have stored for the `cust_id` variable when the request runs. If `cust_id` is currently `3`, the request will be sent to the following URL including query parameter:
+Alternatively, a request body that accesses a variable by wrapping its reference in double-qoutes:
```
SueSmith: Thanks again for editing this - sorry to be a pain but would you mind using the text I mentioned "Alternatively, you could have a request body..." just so that it's a complete sentence? If you aren't keen on that something like "Alternatively, a request body can access a variable by wrapping its reference in double-quotes:" would work.. 😬

jyoung9: My goodness, I didn't even see that that's how I left it..

SueSmith:
@jyoung9 Amazing, thanks so much for adding this, the page gets tons of hits so your example will help lots of people.. 🎊 🎉 🚀

SueSmith:
@jyoung9 just to let you know your contribution is now on the live site https://learning.postman.com/docs/postman/variables-and-environments/variables/#accessing-variables 🙌

PR #2197, opened 2020-02-21
Added a new section to "Run in Postman API"

JJGothard:
Made some language changes and added a section to the bottom about using the index for multiple buttons in on a single page.

### DIFF
--- a/legacy-documentation/v5/postman_for_publishers/run_button/run_button_API.md
+++ b/legacy-documentation/v5/postman_for_publishers/run_button/run_button_API.md
@@ -5,6 +5,131 @@ warning: false
 
 ---
 
+Through the `_pm()` method, you can access your Run in Postman button's API and change the button's behavior.
+
+> The `_pm()` API is not available for the static version of the Run in Postman button.
+
+## Create a new environment
+
+Use the `env.create` method to create a new environment:
+
+```javascript
+_pm('env.create', 'environment_name', {key: value});
+```
+
+For example, if you need to create a new environment using API keys entered by your user, you can do something like this when the Run in Postman button is clicked:
+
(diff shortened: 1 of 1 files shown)


SueSmith:
@JJGothard ok 😬:
* this is still in the wrong directory, if you check my notes from the last time 
* you really need to run the site locally so you can see your changes in the page (which will only work if they're in the right directory)
* you seem to have inserted page content in and above the yaml frontmatter, the frontmatter has to be at the top of the file, if you have a look at some of the current .md files and compare to how they render in the site that should help you understand the structure (the frontmatter has --- before and after it)
* it looks like there are duplicated sections in the page 
  * if you can make your edits inline instead of inserting a completely new section, that will let us see at a glance what language changes you've made (in the diff) - am holding off on providing feedback on the language for that reason as it isn't clear what was already in there and what you've changed
* the rip functionality description is still a little off, the update isn't about segregating the page, it's segregating environments for buttons within a page - for the moment it might be best just to copy in the relevant section from the internal readme I sent you and tidy up the language a little (no need to understand the code at any significant level for the moment, however in order to effectively rework this type of thing you would essentially need to try it out yourself - but that's for another time 🙃)
* this is nbd but for future reference we do try to use [gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) for branch naming - if you check out some of myself and @claireriley's recent prs you'll see some examples

JJGothard:
Yeah, I have run it locally: this was my attempted work-around for the
authentication issue I'm still having.

It inserted funky cos I basically tried to copy/paste the new content,
which is why the other sections are funky too.

Re: Functionality: will rewrite.

PR #2198, opened 2020-02-23
Panel was written pane.

ibadeeCodes:
Just a small spelling mistake.


SueSmith:
@ibadeeCodes thanks so much for your interest in the Postman docs! I'm not sure this is actually a mistake, "pane" is a used term (e.g. https://ux.stackexchange.com/questions/124588/whats-the-difference-between-a-panel-and-pane) however I do wonder if we could word this sentence more clearly in general..? maybe if we say something like "The JSON response will appear below the request detail." 🤔

PR #2199, opened 2020-02-24
Push content and header button updates to prod

SueSmith:


SueSmith:
@ch264 added a few extra bits to develop since opening this pr, some things that are in prod already, if you get a minute to check there isn't anything daft in here so we can get these updates published that would be magic.. 🙌

Issue #1725, opened 2019-07-30
Observations on the Postman Sandbox API Reference Page


DannyDainton:
The dynamic variables were added to the page as part of this PR:

https://github.com/postmanlabs/postman-docs/pull/1717

I noticed a couple of areas that we might need to amend:

The table heading `Decription` is incorrect and should be changed to `Description`, this impacts each table apart from the variables listed under the `Common` section.

The section heading `Text, numbers and colors` is consistent with the rest of the section heading, these Capitalize the first letter of all the words.

The formatting of the tables on the page looks a bit off and could probably do with straightening, up if possible. 

We add the line _Note that dynamic variables cannot be used in the Sandbox. You can only use them in the {{..}} format in the request URL / headers / body._  in the older Postman Sandbox page to let user know _where_ they can use these variables, would we want to include this/a reworded version of this, at the top of the `Dynamic variables` section? 

Would we benefit from having a short code block on the page, using the dynamic variables in a request body for example?

```json
{
	"contacts": [
		{
			"id": "{{$guid}}",
			"fullName": "{{$randomFullName}}",
			"emailAddress": "{{$randomEmail}}",
			"contactPhoneNumber": "{{$randomPhoneNumber}}",
			"dateOfBirth": "{{$randomDatePast}}"
		}
	]
}
```

kaustavdm:
Good catch about the example and the note. I will push a PR to fix these soon. Thanks @DannyDainton!

PR #2205, opened 2020-02-25
Fixed typos in the dynamic variable table

DannyDainton:
Addresses some of the points raised in this issue - https://github.com/postmanlabs/postman-docs/issues/1725


src/pages/docs/postman/scripts/postman-sandbox-api-reference.md
```diff
@@ -451,9 +451,9 @@ The following is a list of dynamic variables whose values are randomly generated
 |                           |                                               | `"53151b27-034f-45a0-9f0a-d7b6075b67d0"`   |
 |                           |                                               | `"727131a2-2717-44ad-ab02-006587e947dc"`   |
 
-### Text, numbers and colors
+### Text, Numbers and Colors
```
SueSmith: @DannyDainton thank you very much for this - only issue is the headings, we've been sticking with sentence case so if you don't mind removing the capitalization from this one and the next one that would be magic 😬

PR #2206, opened 2020-02-25
Added ref to max number of api blogpost

arlemi:
Since the notion of "Max # of APIs" wasn't clear, we've published a blog post about it and we thought it'd be a good idea to add it to the LC too.

Ref: https://blog.postman.com/2020/01/24/what-does-max-of-apis-mean-in-postman/

### DIFF
--- a/src/pages/docs/postman/design-and-develop-apis/the-api-workflow.md
+++ b/src/pages/docs/postman/design-and-develop-apis/the-api-workflow.md
@@ -51,6 +51,8 @@ To access the API Builder, open __APIs__ from the left sidebar in the Postman ap
 
 Click __Create an API__ or __+ New API__—_you will need to be signed into your Postman account_.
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/postman/design-and-develop-apis/the-api-workflow.md
```diff
@@ -51,6 +51,8 @@ To access the API Builder, open __APIs__ from the left sidebar in the Postman ap
 
 Click __Create an API__ or __+ New API__—_you will need to be signed into your Postman account_.
 
+> The number of APIs you are allowed to create is based on your plan. Read more about it [on our blog](https://blog.postman.com/2020/01/24/what-does-max-of-apis-mean-in-postman/).
```
SueSmith: @arlemi tending to keep blog links to the sidebar - think we could include it in the frontmatter instead of in the page body?

arlemi:
TIL what all these links in frontmatter are actually used for. 🤦🏻‍♂️

PR #2234, opened 2020-03-09
Invalid chars docs update

SueSmith:
K have added some very basic details on both invalid chars and url processing based on what I'm seeing in beta right now (anything else can add no probs). Thinking once we have a blog post I can link to it from the sidebar and maybe build a little more detail into the doc..


SueSmith:
Just a note that I've split this into two so that we can get url encoding updates published since it's in prod..

SueSmith:
@shamasis k moved to the troubleshooting section only, updated with the non-whatwg version of the encoding section (same as other branch), added note re right-click but not gone into much detail because it's in the parameters section already..

PR #2236, opened 2020-03-09
URL processing

SueSmith:
Split url encoding and invalid chars into two separate prs - this one is for updates that are in prod already, invalid chars one can wait until release..


SueSmith:
@shamasis got it, was not sure about whether to mention it explicitly or not but will reword

SueSmith:
@shamasis ok reworded and added more detail from your doc - please let me know if any of it is misleading.. 😬

SueSmith:
@claireriley could you have a wee pass through this one for me too? Is in prod so hopefully we can get it published soon.. 😬

PR #2239, opened 2020-03-10
Clarify deletion vs removal of an element notice

lechiffre10:


SueSmith:
Thanks so much for taking the time to add this @lechiffre10! 🎉 

Definitely something we need people to be aware of.. 🚀

PR #2243, opened 2020-03-11
fix video width in mobile view

### LINKED ISSUE #2172: adjust landing page video width
The landing page video needs to be adjusted in mobile view to be the same width as the cards above.

Current state: 
Image from iOS

VibhorCodecianGupta:
fixes #2172 

@ch264 I wrote the media queries in `.scss` file entirely and removed the inline `style` attribute from the JSX tags. Would this be an acceptable approach?

### DIFF
--- a/src/pages/index.jsx
+++ b/src/pages/index.jsx
@@ -97,7 +97,7 @@ The Postman Learning Center has all of the resources you
       </div>
     </section>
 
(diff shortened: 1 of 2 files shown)


ch264:
Hi @VibhorCodecianGupta, Thank you very much for contributing to our Learning Center. This looks great.

PR #2263, opened 2020-03-18
Publish api spec, monitoring docs, live preview, invalid chars, conflict ui docs

SueSmith:
@ch264 couple of docs updates if that's cool


SueSmith:
@ch264 I pushed a few more doc updates to develop - live preview, invalid chars, and conflict ui all of which were released today so if you could give this a wee check over so we can get these updated that would be magic 🙌

SueSmith:
@ch264 something weird is going on in the beta site, certain pages are getting stuck in some sort of refresh loop e.g. am seeing it in https://learning.postman-beta.com/docs/administration/billing/ _I think_ esp if I'm not logged into Postman 😱

SueSmith:
It's only happening on certain pages - the billing one above wasn't actually changed in this pr.. 🤔

SueSmith:
Also happening on this page which _was_ updated, having a hard time finding anything they have in common.. https://learning.postman-beta.com/docs/postman/workspaces/conflicts/ runs fine locally

SueSmith:
...and now it seems to have gone back to normal 😬

ch264:
Looking into this now

tristandenyer:
The Network tab seems to stop at the Algolia call and refresh.

ch264:
The issue is most likely with a redirect. we are looking into the migrated Lambda on AWS

tristandenyer:
I noticed that the production site page works fine, no redirects: https://learning.postman.com/docs/administration/billing/

If this is an issue with the Lambda, and Prod works but Beta does not, then is it safe to say that pushing to Prod should not be an issue? The Lambda is not part of the push to beta or prod—it's outside that pipeline. Meaning, if Prod is woring and beta is not, I would guess that pushing to Prod should NOT get corrupted and not exhibit the same behavior. Right @ch264 and @Ryan-McBride ?

SueSmith:
If we are going ahead and merging to master I have one more pr I'd like to get merged to develop first, was holding off in case it messed up the troubleshooting effort here..

PR #2267, opened 2020-03-23
Delete vs remove info

### LINKED ISSUE #2040: The Difference between Remove Collection and the Delete Collection
It would be great to have the following as a Learning Center doc, as we have a few support tickets that are related to the subject matter. 

*The Difference between Remove Collection and the Delete Collection*

SueSmith:
Fixes #2040

@asi-postman just added a note, hoping it looks ok!


SueSmith:
@claireriley could you give this one a wee once over for me too?

PR #2276, opened 2020-03-25
Updated URL

tristandenyer:
Original URL was using http, not https. And upon further investigation, the URL it was pointing to didn't address the 'static IP option' for that line.


SueSmith:
@tristandenyer I believe this is the link we were given by the pricing team for enquiries on this because you need to request it

ch264:
@tristandenyer can we get an update on this PR?

tristandenyer:
I'm reaching out to Support about this.

tristandenyer:
There is now a lengthy debate happening on Slack. Will update when that gets resolved.

PR #2295, opened 2020-04-03
Added paragraph about Presence

JJGothard:
Added a paragraph about Presence isn "Viewing workspace activity." Added scheenshot of an active workspace member. cc @sankalp0o . Just added two lines to this doc about presence.


SueSmith:
@JJGothard hey can we keep everything on this feature to a single pull request? Everything you commit to the same branch will be added to a single PR even if you commit after opening the PR.

PR #2300, opened 2020-04-03
Chore: fix typo in "gettting started"

viniciusarre:


SueSmith:
Thanks for the fix @viniciusarre! 🙌

Issue #2349, opened 2020-04-23
Typo in Design And Develop APIs > The API Workflow


louisfisch:
In the Next Steps section of the Design And Develop APIs > The API Workflow documentation there is a list of links and the last one reads "Validating elemenst against schema". I believe it should be "Validating elements against schema".

SueSmith:
Thanks for reporting @louisfisch! This might be a good first issue for someone who wants to try opening a pull request to contribute to the docs so have added a label for that. 🙂

louisfisch:
My pleasure. Actually I opened a pull request to fix this issue. But I think GitHub is experiencing some issues and the checks can't run.

louisfisch:
I am closing this issue because the merged pull request #2350 fixed this issue.

SueSmith:
Oops hadn't clicked we weren't closing this one automatically - thanks @louisfisch!

PR #2350, opened 2020-04-23
Correct typo in The API Wokflow docs

louisfisch:
Fix the issue #2349

### DIFF
--- a/src/pages/docs/postman/design-and-develop-apis/the-api-workflow.md
+++ b/src/pages/docs/postman/design-and-develop-apis/the-api-workflow.md
@@ -17,7 +17,7 @@ contextual_links:
     name: "Create APIs directly within the Postman app"
     url: "https://blog.postman.com/2019/05/17/postman-7-1-create-apis-directly-within-the-postman-app/"
   - type: link
-    name: "What does “Max # of APIs” mean in Postman?"
+    name: "What does 'Max # of APIs' mean in Postman?"
     url: "https://blog.postman.com/2020/01/24/what-does-max-of-apis-mean-in-postman/"
   - type: section
     name: "Next Steps"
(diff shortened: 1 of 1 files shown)


SueSmith:
That was quick! Thanks so much @louisfisch 🎉 🚀 🏆

SueSmith:
Not sure why the check is failing but looks like an issue on our end so we'll get it sorted..

SueSmith:
Also something weird is happening with my notifications, getting them in the wrong order lol, GitHub having a few issues I think 🙈

louisfisch:
You're welcome. I think it's on the GitHub's end though https://www.githubstatus.com/incidents/2y6v4ltq26g7

SueSmith:
Ah great thanks hadn't seen this. Will try again later!

louisfisch:
Is there something I can do so the build can pass? I don't fully understand why it is failing. I thought it was due to the GitHub incident but it keeps failing even though the incident was resolved.

SueSmith:
@louisfisch hm I was wondering why it's still failing too, other prs are now passing.. anyway it shouldn't be on your end, you've literally changed one line of text..

Paging @ch264 who I believe set these checks up and am hoping has more of a clue than me into what they actually do.. 😬

ch264:
@louisfisch Thanks so much for contributing 🚀 I have not experienced this error message before. This might be a rebase or clone issue resulting in an issue with GitHub where is does not recognize the commit.

I would probably recommend to create a new branch and create a new pull request with the fix (possibly cloning down the repo again and start afresh)? Sorry, I will research some more into this issue so for I would recommend a fresh start for the PR 😄

SueSmith:
@louisfisch commited a daft little change to your pr to see if that somehow bypassed the weirdness and it seems to have worked lol 🙃

PR #2360, opened 2020-04-29
FAQ style

SueSmith:


DOCS_STYLE_GUIDE.md
```diff
@@ -334,6 +337,38 @@ Don't display a raw URL in text, unless seeing the URL is essential to the learn
...
+---
+
+:thumbsup::white_check_mark: Contact us at sales@postman.com and we will...
+
+---
+
+Don't include subheadings in your FAQ content. These appear as page sections within a topic page, so to maintain consistent navigation don't include any subheadings inside your FAQ.
+
+You can make time-specific references in FAQs—if the content relates to a temporary problem.
+
```
claireriley, line 355: Maybe we can add a note here that the person authoring the time-specific FAQ is the first person in charge of removing it once the issue is resolved if that makes sense? or we could create some sort of spreadsheet with links to all temporary faqs to keep track of resolved ones

DOCS_STYLE_GUIDE.md
```diff
@@ -334,6 +337,38 @@ Don't display a raw URL in text, unless seeing the URL is essential to the learn
...
+---
+
+:thumbsup::white_check_mark: Contact us at sales@postman.com and we will...
+
+---
+
+Don't include subheadings in your FAQ content. These appear as page sections within a topic page, so to maintain consistent navigation don't include any subheadings inside your FAQ.
+
+You can make time-specific references in FAQs—if the content relates to a temporary problem.
```
claireriley, line 354: Maybe this could be with the stipulation that the issue has to be one that we anticipate to last a minimum of x days

DOCS_STYLE_GUIDE.md
claireriley: I suppose if we anticipate non-postman authors the first part wouldn't work, although I'd imagine they'd probably tend towards writing more permanent faqs

DOCS_STYLE_GUIDE.md
claireriley: (not sure if this would necessarily be included in the style guide though, more a part of the decision-making process)

DOCS_STYLE_GUIDE.md
SueSmith: Aye good call re removing.. Yeah I would guess so too re external contributors - I wonder if I should also say if you aren't sure where to add your contribution open an issue and we can discuss..

jkonrath-postman:
Closing this because we're going in a different direction on this. See TW-4 internally or ping me for details.

PR #2366, opened 2020-05-05
Update docs for new importers release

claireriley:
Docs for upcoming 1.0 Importers release. 

Includes:
* Updated/refreshed content
* New page name
* Redirects
* Updated links

PR #2383, opened 2020-05-13
Update fixes

claireriley:
@SueSmith I think this addresses almost all of the fixes you suggested [here](https://github.com/postmanlabs/postman-docs/pull/2366). 

The one thing I didn't change yet is the list of formats. You had suggested a table or a different layout since the current format suggests it's leading users to another part of our docs. I was thinking that too, but the table looks odd to me as it's a single column unless I split out the links and it's so brief.. Screen Shot 2020-05-13 at 3 40 10 PM
I'm wondering if there's an alternative?

### DIFF
--- a/src/pages/docs/postman/collection-runs/integration-with-travis.md
+++ b/src/pages/docs/postman/collection-runs/integration-with-travis.md
@@ -70,7 +70,7 @@ Let's learn more about integration with Travis:
 
 ## Hooking up Postman to Travis CI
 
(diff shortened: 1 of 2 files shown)


claireriley:
@SueSmith I also sent a message to support regarding the checks not running, haven't seen that before.

SueSmith:
@claireriley aye it looks a bit weird as a table, it's cool as is anyway, thinking we should probably style external links differently across the board in any case..

PR #2400, opened 2020-05-19
Develop merge master

oswish:
_Branched from `develop`_, this merges `master` into `develop`, updates a snapshot, & does lint fixes.

w/ this branch `develop_merge-master`:

* It's concerning that commits that are in `master` (PROD), are not in `develop`, ATM; commits that harken back as far as Oct 30, 2019; _though only 9 files changed_
* confirmed `jest` runs successfully
* confirmed `npm run test:lint` runs clean
* crawling a locally running LC has no 404s

### DIFF
--- a/src/components/CookieAlert.jsx
+++ b/src/components/CookieAlert.jsx
@@ -36,7 +36,7 @@ const CookieAlert = () => (
     >
       This website uses cookies to ensure you get the best experience on our website.
       {' '}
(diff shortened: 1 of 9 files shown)


oswish:
@tristandenyer, @SueSmith; though, I'm realizing all those old commits are only merge commits, which explains why it's only "Files changed (9)".

PR #2409, opened 2020-05-26
Add recent blog posts

SueSmith:
Just added some recent blog posts that followed feature release to docs sidebar (need to build this into the release docs flow..)

cc @kinlane @loopDelicious

### DIFF
--- a/src/pages/docs/integrations/github.md
+++ b/src/pages/docs/integrations/github.md
@@ -14,8 +14,8 @@ contextual_links:
   - type: subtitle
     name: "Related Blog posts"
   - type: link
-    name: "Backup and Sync your Postman Collections on GitHub"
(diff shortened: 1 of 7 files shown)


SueSmith:
Holding off here for updated version of codegen post as per @loopDelicious

loopDelicious:
> Holding off here for updated version of codegen post as per @loopDelicious

@SueSmith re: code gen blog post, the title has been updated, but I'm still waiting for the body of the post to be pushed to production. 

Since there's currently no way to preview changes for the blog, might take a few iterations to format the post for legibility, maybe a few days for each push to production, which could take a while.

loopDelicious:
@SueSmith looks like the copy has been updated! 

Code gen blog is good to go from my end. Since it takes so long to preview updates, I'm planning to leave the images as illegible thumbnails until we update the blog publishing process 😬

Issue #2410, opened 2020-05-26
Typo in sidebar link


SueSmith:
https://learning.postman.com/docs/postman/collections/examples/

"Ingur" > "Imgur"

SueSmith:
Typo is in the sidebar (which is defined in the yaml at the top of the md file).

HarshaVardhan16:
Hey,
I am interested in working. Guide me, so that I can understand the project better.

Issue #2357, opened 2020-04-27
404 image


ch264:
https://assets.postman.com/postman-docs/WS-certificates.gif

should be 

https://assets.postman.com/postman-docs/WS-certificates.png

SueSmith:
Image is used in this page/section https://learning.postman.com/docs/postman/design-and-develop-apis/validating-elements-against-schema/#validating-elements

HarshaVardhan16:
Hey,
I am interested in working. Guide me, so that I can understand the project better.

PR #2439, opened 2020-06-03
Update example.md

HarshaVardhan16:
Changed "Ingur" > "Imgur"


SueSmith:
Hi there, thanks so much for contributing, we actually have an assignee working on two of these issues already (#2410 and #2357) 🙃 if you could stick to issues that don't have an assignee that would be great 🙌

SueSmith:
Same here, thank you again 🙌

PR #2450, opened 2020-06-09
New Algolia Search UI 

ch264:
We are streamlining the search experience across our web properties.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,7 +1,11 @@
 const queries = require('./src/utils/algolia');
 
+// require('dotenv').config({
+//   path: `.env.${process.env.GATSBY_ACTIVE_ENV}`,
+// });
+
 require('dotenv').config({
-  path: `.env.${process.env.GATSBY_ACTIVE_ENV}`,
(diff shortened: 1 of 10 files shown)


SueSmith:
Thanks @ch264! Few wee queries but def just enhancements:
* Could the text sections be word wrapped so that words aren't split across lines?
* Could we crop the video text results?
Screenshot 2020-06-10 at 09 24 13

* For the LC results could we have some way of enclosing the text portion so it doesn't look so disembodied (can't think of a better word 🤪) e.g. ellipsis (...) or something at the start / end?
Screenshot 2020-06-10 at 09 32 10

(See Algolia's own search results for an example of what I'm on about re ellipsis:)
Screenshot 2020-06-10 at 09 39 36

arlemi:
@ch264 Are we losing the ability to get search results straight away with this new implementation? 

With the current one, as soon as I start typing I get suggestions: 
searchLC

With the new one it seems I need to trigger a search to then be redirected to a result page:
searchBlog

Can we disable suggestions from the browser too as it was before? 
image

Now:
image

PR #2452, opened 2020-06-10
push to prod

ch264:


ch264:
checking now

SueSmith:
I'm also not getting relevant search results from entered terms, only relevant ones I'm getting are when I click a trending search.

arlemi:
Re: 
> @arlemi the new search UI on /search still has the ability to get search results on key stroke.

I can't seem to get it to work on `/search` either. Not being able to search on keystroke is a regression from the last implementation we had which will be disturbing to users and adds more steps to get to the results. 

If we can't revert to search on keystroke everywhere, can we at least make it obvious to the users that they now need to press enter to trigger a search? or add a search button?

tristandenyer:
@arlemi On the `/search` page, have you tried typing in the blue text? It should give you results as you type. Here:
Screen Shot 2020-06-11 at 3 47 00 PM Perhaps we make the 2nd search field more obvious (but not just duplicative of the one above it so as to not have 2 identical boxes on top of each other.)

The Search across the sites is developing toward supporting federated search—pulling in results from multiple sites. Showing numerous results as you type across 3 or more sites is not possible in the previous, small dropdown. To support the more robust results, we need a page.

This is a departure from before, and an iteration toward helping users find what they are looking for—which may be on another Postman site. It is also heavily inspired by existing search experiences on sites like IKEA, Parachute, algolia.com and more, keeping in line with making search more like other common approaches.

I will add a ticket to incorporate a Search button to see if this helps alleviate the friction. Design is working on the redesign as well, and will be sure they get this feedback.

ch264:
@Ryan-McBride and Me are looking into the left nav bar being broken.

arlemi:
@tristandenyer The three examples you mentioned allow search on keystroke from any page and do not require an additional action to access search results. Algolia loads a new page as soon as a user clicks in the search bar which is transparent to them. 

I see the benefits of pulling results from different properties, I'm unsure why we can't have some of them in the area we're opening under the search bar, for example: 
image

This would mean no regression and improving the search feature to show results from all properties, everyone wins.

ch264:
@tristandenyer @SueSmith https://github.com/postmanlabs/postman-docs/pull/2463 solves the left-nav bar not reacting to clicks.

SueSmith:
@ch264 Still not seeing relevant search results in beta btw fyi 🙃 (only seems to return relevant ones on clicking a trending search, but not from a term entered into the search manually)

ch264:
@SueSmith can you please give me an example? I cannot reproduce a difference in results.  I have one example for variables here. 

This is the result list for a current search for 'variables':
Screen Shot 2020-06-17 at 4 14 00 PM

This is the result list for a search for 'variables' with the new component:
Screen Shot 2020-06-17 at 4 13 40 PM


The API call and the index remain the same. What is changing in the new search UI is that there is one more additional search result to before and a component that calls real-time trending search terms

Ryan-McBride:
I can reproduce Sue's error in beta as well.
Steps:
go to blog.postman-beta.com
click the search bar
enter 'variables'
no results are shown, just trending searches
[image: Annotation 2020-06-17 163321.png]

SueSmith:
If I search in any way other than clicking on the trending list I get no relevant results, just the results that appear by default at /search. If I type "variable" and hit enter I don't get the right results, I only get them when I click "variable" in the trending searches. Try entering any of these for example "monitors" "certificates" "mock" all of which have pages with the terms in the title - if you search these in the prod site you'll see the relevant results.

SueSmith:
This is what I'm seeing on https://learning.postman-beta.com/ btw, @ch264 looks like your screengrab is localhost..

FYI we will have docs updates to publish for Monday's release so if this isn't resolved by then can we revert it so that we don't break the search when we publish.. 😬 🙏

ch264:
@SueSmith @Ryan-McBride The Modal does not show on-type search, which is the expected behaviour. On enter, the search will be sent to Algolia and the results will be displayed on /search. 

I can see that there is a difference in search results. 
When entering a search term and clicking enter - the same results are shown as now.
When clicking on a trending search - other results are shown. 
I will contact Algolia and get an explanation. 

In the meantime, @SueSmith of course we can revert the search for the release on Monday should there be more open questions.

SueSmith:
@ch264 this isn't what I'm seeing on the beta site btw:
> When entering a search term and clicking enter - the same results are shown as now.

I _only_ see relevant results on clicking a trending search. The results when I type and enter are not relevant at all.

ch264:
@SueSmith @arlemi @Ryan-McBride 
I found what is causing the difference in search results. I have removed federated search for now so that you can release on Monday

PR #2461, opened 2020-06-15
revert search in nav bar - breaking left nav

ch264:


ch264:
I found the bug. See PR #2463

PR #2473, opened 2020-06-18
feature: add hover to left navbar link

### LINKED ISSUE #2469: Left navbar is missing hover states
When a user hovers over the dark gray text links in left nav, the label should turn to orange (`color: #FF6C37;`)

Also, add `display: block;` to the href so that you don't have to only click on the label, but be able to click anywhere along the horizontal axis of that label, making it a lot easier to engage with.

junagao:
resolves #2469


ch264:
@junagao Thank you very much for contributing 🚀 👏 Your changes will go to prod in the next release

PR #2476, opened 2020-06-19
Dropdown removal

SueSmith:
@claireriley sorry for the avalanche of review requests, if they don't all get reviewed nbd will just publish on mon and we can refine later.. 😭 this one is for the new button which is having the dropdown list removed, the new button doc is getting binned in the restructure (and gets hardly any views) so have kept this to the absolute bare minimum, screengrabs of it are all over the place so will need to be updated in the longer term anyway 🙃


SueSmith:
..and @arlemi was daft enough to volunteer for some of these reviews lol 🙌

Issue #2444, opened 2020-06-04
FAQs Let’s Encrypt SSL certificates


cyrielle-hada:
**Summary**
Surface Let’s Encrypt SSL certificates expiration information.
- When does it expire?
- What will happen once the certificate expires?
- What to do?

**Support reported**
```
We have a couple of published postman collections. I've noticed that the Postman-managed SSLs from Lets Encrypt will both expire next week. 
- Do they auto-renew? 
- When does that happen normally? 
- Is there any action required from our side?
```
**Learning Center location**
https://learning.postman.com/docs/postman/sending-api-requests/certificates/

**Suggestion**
Add an FAQ to the Certificate doc page and address the following questions.

- **Do I need to renew the certificate?** 
Certificates are renewed automatically, no manual steps are required.

- **How long are these SSL Certificates valid for?**
Certificates are generated with a 90-day expiry date.

- **When does Let’s Encrypt renew SSL certificates for published documentation?**
The certificate gets renewed 7 days before it expires.

fakela:
I would love to work on this

PR #2514, opened 2020-07-09
documentation on FAQs Let’s Encrypt SSL certificates

fakela:
This PR is to fix documentation on FAQs Let’s Encrypt SSL certificates #2444

### DIFF
--- a/src/pages/docs/sending-requests/certificates.md
+++ b/src/pages/docs/sending-requests/certificates.md
@@ -22,8 +22,6 @@ warning: false
 
 Postman provides a way to view and set SSL certificates on a per domain basis.
 
-> Certificates renew automatically—you do not need to carry out any manual steps. When a certificate is generated it has a 90 day expiry date and will renew seven days before it expires
-
 To manage your client certificates, click the wrench icon on the right side of the header toolbar, choose "Settings", and select the **Certificates** tab.
 
 [![certificates tab](https://assets.postman.com/postman-docs/WS-certificates.png)](https://assets.postman.com/postman-docs/WS-certificates.png)
@@ -69,6 +67,8 @@ To remove a certificate, use the **Remove** link next to the certificate under 
 ## Editing a Certificate
 
 You cannot edit a certificate after it has been created. To make changes to it, you will need to remove the certificate and create a new one.
+
+> Let's Encrypt SSL certificates renew automatically—you do not need to carry out any manual steps. When a certificate is generated it has a 90 day expiry date and will renew seven days before it expires
 
 ## Certificate data
 
--- a/src/pages/docs/sending-requests/certificates.md
+++ b/src/pages/docs/sending-requests/certificates.md
@@ -22,6 +22,8 @@ warning: false
 
 Postman provides a way to view and set SSL certificates on a per domain basis.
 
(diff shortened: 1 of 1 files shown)


SueSmith:
Hi @fakela! Thanks so much for the PRs. 🙌 

So we actually have a docs style guide: https://github.com/postmanlabs/postman-docs/blob/develop/DOCS_STYLE_GUIDE.md

Typically in a case like this we would build the information into the body of the text in the doc rather than adding it as an faq section (in paragraphs and sentences rather than in list style). We also don't use first person in our docs by default ("I" in this case).

Please give me a shout if I can provide any other info! I'll hold off on your other PRs for now but your contribution is much appreciated. 🙂

fakela:
@SueSmith  thanks for your comment and suggestion👍 . Based on your suggestion I came up with this sentence as it covers the answers given in the faq, would it be okay under **using a certificate** section

Note: Certificates are valid for 90 days and are renewed automatically 7 days before it expires.

SueSmith:
Thanks for following up @fakela - few notes, please don't feel obliged to action all this, just to give you a sense of how we edit content in the docs team:

* I think a note in that section would work 👍 
* For notes we normally use markdown blockquotes (`>`) rather than including "note" in the text (see style guide)
* The sentence you have would need "it" changed to "they" because the first part of the sentence is plural, however I'd rework it anyway, as follows:
 * We try not to use passive language so I'd reword to use active phrases, and would split it into a couple of sentences to highlight the automatic part. We also spell out numbers up to ten.. e.g. something like this: "Certificates renew automatically—you do not need to carry out any manual steps. When a certificate is generated it has a 90 day expiry date and will renew seven days before it expires."
* I think we would also need to indicate that this information is specifically for Postman managed Let’s Encrypt SSL certificates since the certificates doc is more general if that makes sense

Again please don't feel obliged to go through all these, for a community contribution we would typically make some of the edits within the team but I know you were interested in how we work on the docs.. Thanks again! 🙂

fakela:
> Thanks for following up @fakela - few notes, please don't feel obliged to action all this, just to give you a sense of how we edit content in the docs team:
> 
> * I think a note in that section would work 👍
> * For notes we normally use markdown blockquotes (`>`) rather than including "note" in the text (see style guide)
> * The sentence you have would need "it" changed to "they" because the first part of the sentence is plural, however, I'd rework it anyway, as follows:
> * We try not to use passive language so I'd reword to use active phrases and would split it into a couple of sentences to highlight the automatic part. We also spell out numbers up to ten.. e.g. something like this: "Certificates renew automatically—you do not need to carry out any manual steps. When a certificate is generated it has a 90 day expiry date and will renew seven days before it expires."
> * I think we would also need to indicate that this information is specifically for Postman managed Let’s Encrypt SSL certificates since the certificates doc is more general if that makes sense
> 
> Again please don't feel obliged to go through all these, for a community contribution we would typically make some of the edits within the team but I know you were interested in how we work on the docs.. Thanks again! 🙂


@SueSmith Thanks for the detailed explanation. I would use this "Certificates renew automatically—you do not need to carry out any manual steps. When a certificate is generated it has a 90 day expiry date and will renew seven days before it expires" to update the docs.

fakela:
@SueSmith Please review

SueSmith (approved):
moved this and added a little detail..

PR #2547, opened 2020-07-16
Update gitlab.md

claireriley:
Updating for clarification—GitLab is a one-way backup, whereas "sync" may suggest bidirectional (as it does in the case of GitHub)


claireriley:
cc @SueSmith

Issue #2509, opened 2020-07-08
Include header text in search index


loopDelicious:
Looks like we might not be indexing headers in search anymore.

Searching for `console`, displays the Postman console as the last result, even though it's our most popular console, and the page includes "console" in a header.

![Screen Shot 2020-07-08 at 2 53 19 PM](https://user-images.githubusercontent.com/17693714/86977612-d8053680-c131-11ea-8818-07c7856772f5.png)

Also, there's max results displayed, so if we add another page with the console, the Postman console won't be displayed as a result.

ch264:
@loopDelicious I just checked and 'titles' are being indexed by Algolia. This has not changed. 
In the legacy documents there used to be a doc with that had the word 'Console' in it. This doc has now been renamed. 

For max results, this is a design choice and the design team will have to provide us with a new design. There is an open ticket in Jira for this. I will send you the link.

I can look into popularity ranking and check out display of hits again. I agree that the most popular post doc should be on top.

loopDelicious:
Also the links in the Contents section of this page have an extra `/` in them so they're not linking to the sections indicated.

https://learning.postman.com/docs/sending-requests/troubleshooting-api-requests/#contents

ch264:
@loopDelicious I am not sure I understand. Is this related to Algolia search? 
 When I open the links in the content section, they go to the correct section of the page.
<img width="1111" alt="Screen Shot 2020-07-14 at 1 59 39 PM" src="https://user-images.githubusercontent.com/36343528/87475814-463a7500-c5da-11ea-88e2-f1f92d3ac0eb.png">

loopDelicious:
Here's another example - searching `switch accounts` yields limited results, but not a header `Switching between accounts` on [this page](https://learning.postman.com/docs/getting-started/postman-account/#switching-between-accounts).

![Screen Shot 2020-07-15 at 2 09 27 PM](https://user-images.githubusercontent.com/17693714/87596386-dc3ad200-c6a4-11ea-8944-7cc6a27733e0.png)

loopDelicious:
> @loopDelicious I am not sure I understand. Is this related to Algolia search?
> When I open the links in the content section, they go to the correct section of the page.

Not sure if someone fixed it? Here's another example of links in the contents section that have the extra `/`. Not sure if this was human error, or a migration issue.

[...]

PR #2561, opened 2020-07-24
Search subheadings

SueSmith:
Re #2509 this adds headings as a searchable attribute in Algolia

I have just tested this out very briefly in the demo site I had used for the restructure but here's what I've got set up in my own test Algolia index:

* increased typo tolerance to five chars (fixes "fork" query issue)
* selected searchable attributes as follows - to rank title, headings among other detail

attributes
ranking

When I search in the **Browse** tab in Algolia I get better results (e.g. "console" returns troubleshooting page first and top five are all more relevant, "variables" returns dynamic and using variables / environments pages etc), however:

* I'm not seeing the same results in my live demo site which may be on account of me missing some part of the build process (ran out of time to continue investigating this 🙃), but am wondering if there is an issue connecting the updated Algolia site to our live site - @ch264 if you could try a couple of queries in the Algloia **Browse** view and in the live site and see if you get different results that would confirm this..
* Although I'm finding these tweaks addresss most of the problematic queries we've been talking about, there are still some serious limitations to Algolia's algorithm that I think may mean we need to explore other options:
    * @loopDelicious you mentioned "switch account" doesn't return page with "switching accounts" in it, this is because Algolia does not do "stemming" i.e. treating "switch" and "switching" as equivalent.
    * The fact that Algoia does not offer any way to factor frequency of search term occurrence in page is imo always going to return less relevant results first.
    * _It looks like Elasticsearch does both of these so we might want to look into that at some point._

Re including more search results, am seeing `hitsPerPage` in `header.jsx` but held off on changing it in case there are layout implications..

In the meantime if we could try the above tweaks and check that nothing is broken in terms of linking the updated index to the live site that should hopefully be a significant improvement.

### DIFF
--- a/src/utils/algolia.js
+++ b/src/utils/algolia.js
@@ -15,6 +15,9 @@ const pageQuery = `{
   ) {
     edges {
       node {
+        headings(depth: h3) {
+          value
+        }
         frontmatter {
           title
           search_keyword


SueSmith:
Forgot to say that what I had found initially was that the results weren't ordered by ranking elements but just by object id, I think we were maybe not providing enough attributes to rank on..

ch264:
@SueSmith PR looks great. Once it's merged I can add heading to search attribute in Algolia. I have also increased typo tolerance from 4 to 5.

ch264:
@SueSmith I checked and Algolia allows to add synonyms. So if you could send me you most searched for synonyms that Algolia does not handle by default, we can address these manually for now

SueSmith:
Thanks @ch264 please do what you want with this pr, I just opened it to put all this in one place.. Yeah I had seen the synonym bit too but obvs not sustainable to do manually in general..

ch264:
For the change in search results, you can do that by adjusting the number. The component is dynamic so it will adjust to the number of search results you want to display.

Screen Shot 2020-07-24 at 11 58 32 AM

SueSmith:
Great, paging @loopDelicious re how many results are displayed, we can just alter the number.. ^^

ch264:
@SueSmith do you want to merge this PR or shall I create a new one for the headings?

SueSmith:
@ch264 happy to merge if that works!

PR #2584, opened 2020-08-13
[Fix #2575] Start search at 3 characters

### LINKED ISSUE #2575: Display search results when 3 characters are being typed
Currently search results are being displayed on every key stroke. Change Algolia to start searching on every 3rd keystroke

AnujSharma141:
Fix #2575 
Used a ternary expression to change the state accordingly after event.currentTarget.value.length>3 in the onKeyUp event.


ch264:
Hi @AnujSharma141, Thanks a lot for contributing and working on this 🚀  We already have a fix in, sorry I should have noted in the issue that this was solved already. Thanks again 🙌

PR #2614, opened 2020-08-28
Correct Link to API

leemicw:
The API link currently points to www.postman.com instead of api.getpostman.com


SueSmith:
Hi @leemicw! Thanks for the PR. You can actually run collections in newman without using the api url if the collection link is public so the "api." isn't necessarily required - however since that wasn't clear I wonder if we should add a note to indicate it instead..?

leemicw:
@SueSmith I didn't know public collections were available on the www server.  If I remember correctly I got a 404 even after adding the correct api key.  If yall are happy with the link to www feel free to close the PR.

claireriley (approved):
Thanks @leemicw and @sowmyaraj92 !

Issue #2635, opened 2020-09-09
Add direct link flow to accessing Postman API keys


SueSmith:
In this page https://learning.postman.com/docs/developer/intro-api/ we have how to access your api keys via the ui, but you can also access them using this url: https://go.postman.co/settings/me/api-keys so we should add this to the doc

carmelag:
@SueSmith hi! I would like to take this one. Is anyone else working on this? Please lmk

SueSmith:
Hi @carmelag! If you'd like to work on it that would be great 🙌

carmelag:
Ok @SueSmith! Great! Could you please assign the issue to me? I'll let you know when it's done. Thanks for your quick response 🙌

SueSmith:
Thank you for contributing! 🙂

PR #2647, opened 2020-09-22
New direct link for API key creation added on top of alternative dash…

carmelag:
I added the direct link to the API Key page on top of the section "Generating a Postman API Key". The procedure is the same as the dashboard's one, so the pictures previously used for the dashboard procedure are still valid also for the direct link. 

I moved the dashboard procedure at the end of the section, considering it as an alternative way to achieve the purpose of creating a new API Key.

@SueSmith please let me know what you think and if this approach works for you.
Thanks!

### DIFF
--- a/src/pages/docs/developer/intro-api.md
+++ b/src/pages/docs/developer/intro-api.md
@@ -19,21 +19,23 @@ warning: false
 
 ---
 
-The [Postman API](https://docs.api.getpostman.com/) endpoints to help you integrate Postman within your development toolchain.
+The [Postman API](https://docs.api.getpostman.com/) endpoints to help you integrate Postman within your development toolchain.
 
-You can add new collections, update existing collections, update environments, and add and run monitors directly through the API. This enables you to programmatically access data stored in your Postman account.
+You can add new collections, update existing collections, update environments, and add and run monitors directly through the API. This enables you to programmatically access data stored in your Postman account.
 
-You can get started with the API via the **Run in Postman** button at the top of the [Postman API documentation](https://docs.api.getpostman.com/) and use the Postman app to send requests.
+You can get started with the API via the **Run in Postman** button at the top of the [Postman API documentation](https://docs.api.getpostman.com/) and use the Postman app to send requests.
 
 You will need an [API key](#generating-a-postman-api-key) to access the Postman API.
 
 > The Postman API is [rate limited](#rate-limits).
 
-## Generating a Postman API Key
+## Generating a Postman API key
 
-You need a valid API Key to send requests to the Postman API endpoints.
+You need a valid API Key to send requests to the Postman API endpoints.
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/developer/intro-api.md
```diff
@@ -29,13 +29,11 @@ You will need an [API key](#generating-a-postman-api-key) to access the Postman
 
 > The Postman API is [rate limited](#rate-limits).
 
-## Generating a Postman API key
+## Generating a Postman API Key
```
SueSmith, line 32: We tend to use sentence case for subheadings so if you could revert this change that would be great

src/pages/docs/developer/intro-api.md
```diff
@@ -29,13 +29,11 @@ You will need an [API key](#generating-a-postman-api-key) to access the Postman
...
 
 You need a valid API Key to send requests to the Postman API endpoints.
 
-Navigate to the [web dashboard](https://app.getpostman.com/). Select a workspace and open __Integrations__. Locate the Postman API and click __View Details__.
-
-![Workspace Integrations](https://assets.postman.com/postman-docs/workspace-integrations-api.jpg)
+To generate a new API Key go to the [Postman API keys page](https://web.postman.co/settings/me/api-keys).
```
SueSmith, line 36: I think I'd keep the instructions in the same place so that the flow is reliable regardless of how the person chooses to access the key it if that makes sense, the link isn't the preferred route or anything, just one option..

SueSmith (changes requested):
Thank you so much for the contribution @carmelag! 🙌 A couple of wee requests if you don't mind.. 🙏 

Also if you add "Fixes #2635" to your PR description text that will automatically close the issue when we get your great PR merged.. 🙂

src/pages/docs/developer/intro-api.md
carmelag: Hi @SueSmith I will apply the required changes during the weekend. Thanks for your quick feedback.
So the order will stay the same:  "UI approach" and "Direct Link approach" (as an alternative).
Is that correct?

SueSmith: Hi! Yep I think I'd just build it into the existing text, e.g. starting the section with something like:

```
You can generate an API key directly from your [Postman API Keys](https://go.postman.co/settings/me/api-keys) page or by navigating to your dashboard.
```

Then just continuing with the original instructions.. 

(Also I think the link shortcut may have changed 😭 so if you don't mind changing the link from `web.postman` to `go.postman` as I have in this example that would be great.) 

Sound reasonable? Thanks again!

carmelag: Hi @SueSmith I applied the changes as you suggested. Could you please have a look at the final version after last commits? 

Thanks

SueSmith (approved):
LGTM! 👍

SueSmith:
Thanks so much for contributing @carmelag! 🙌 🎉 🚀

Issue #2183, opened 2020-02-13
Request to remove sync screenshot from docs


Raj-postman:
As disabling sync option is removed from the application, can we please remove this particular section from the following document:

https://learning.postman.com/docs/postman/monitors/troubleshooting-monitors/#debugging-local-run-attempts

Screenshot:

<img width="1041" alt="Screenshot 2020-02-13 at 4 22 52 PM" src="https://user-images.githubusercontent.com/47771757/74428395-ebf88500-4e7e-11ea-8ed9-92ac06f47874.png">

Raj-postman:
Not sure if this is being tracked somewhere but this https://github.com/postmanlabs/postman-docs/issues/1697 section needs to checked again.

arlemi:
@Raj-postman I believe @claireriley is currently updating the monitoring docs so this would probably fall into it?

Regarding #1697, the page says deprecated at the top and can't be accessed apart by searching for team library. We need to look into why it still shows in search results though. 🤔

loopDelicious:
@Raj-postman this one has been fixed by the monitoring updates. Feel free to reopen if you have more to add.

Raj-postman:
@loopDelicious I still see the screenshot of disabling sync in that document there:

It should be removed or else it's bound to create confusion 😓 

Current screenshot:

![image](https://user-images.githubusercontent.com/47771757/94174720-48129580-feb3-11ea-9a27-39940854a20a.png)

Actual Settings Modal:

![image](https://user-images.githubusercontent.com/47771757/94174586-20233200-feb3-11ea-9fa5-84263621f3ff.png)

loopDelicious:
Ah, I see it now. Sounds like we just need to remove the bullet and screenshot 👍

PR #2657, opened 2020-09-30
Fix for  #2183

jerry-li-dev:
Removed the incorrect bullet point and screenshot.

https://github.com/postmanlabs/postman-docs/issues/2183

### DIFF
--- a/src/pages/docs/designing-and-developing-your-api/monitoring-your-api/troubleshooting-monitors.md
+++ b/src/pages/docs/designing-and-developing-your-api/monitoring-your-api/troubleshooting-monitors.md
@@ -43,10 +43,6 @@ For example, if the monitor failed because Postman was unable to contact the ser
 
     [![enable sync](https://assets.postman.com/postman-docs/Header_Toolbar_New_Sync.png)](https://assets.postman.com/postman-docs/Header_Toolbar_New_Sync.png)
 
-* You can also force a sync from the **Sync** tab within the **SETTINGS** modal.  
-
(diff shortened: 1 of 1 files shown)


claireriley (approved):
Looks great! You can go ahead and merge 😄 

A good next small issue would be to update this line:
> If a local run passes, [ensure the sync is enabled](/docs/getting-started/syncing/) in the toolbar header of the Postman app to persist any local changes.

I think "enabled" suggests there's still an option to disable (we removed that a while back), so it'd be better to say something along the lines of "ensure sync is working by...". If you're up for it, you can create an issue and a PR for the fix!

jerry-li-dev:
> Looks great! You can go ahead and merge 😄
> 
> A good next small issue would be to update this line:
> 
> > If a local run passes, [ensure the sync is enabled](/docs/getting-started/syncing/) in the toolbar header of the Postman app to persist any local changes.
> 
> I think "enabled" suggests there's still an option to disable (we removed that a while back), so it'd be better to say something along the lines of "ensure sync is working by...". If you're up for it, you can create an issue and a PR for the fix!

Thank you I will make the issue in the morning and take care of it :)

claireriley:
@jerry-li-dev I realized we don't have this noted in the [contribution guidelines](https://github.com/postmanlabs/postman-docs/blob/develop/CONTRIBUTING.md#pull-request-process)—I'm going to add it today—but we'll want to get in the habit of deleting the associated branch following each merge to keep the repo clean! You should see a button in this thread

PR #2702, opened 2020-11-14
Documentation Flow Updates

jerry-li-dev:
Doc: Document your API

new procedures and screenshots

### DIFF
--- a/src/pages/docs/publishing-your-api/viewing-documentation.md
+++ b/src/pages/docs/publishing-your-api/viewing-documentation.md
@@ -56,9 +56,9 @@ You can [view documentation for any collection that has been shared with you](/d
 
 To view documentation for a collection you have access to in Postman, select the collection, click the three dots (...), and choose __View Documentation__. The collection documentation will open within a new tab in the application (both in the browser and desktop application).
 
-![View Docs](https://assets.postman.com/postman-docs/view-docs.png)
+![View Docs](https://assets.postman.com/postman-docs/view-docs-b.jpg)
 
-At the top, you will see the documented collection, the sample code language, and the layout of the documentation. You can change the language and layout.
+At the top, you can select an environment, layout, and sample language.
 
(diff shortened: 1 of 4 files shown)


src/pages/docs/publishing-your-api/viewing-documentation.md
```diff
@@ -80,7 +80,7 @@ If the documentation you are viewing has multiple versions published, you can se
 
 You can view and post comments on your API documentation to collaborate with teammates. When you view the documentation for a collection, you will see any existing comments on it.
 
-![Documentation Comment](https://assets.postman.com/postman-docs/documentation-comment.jpg)
+![Documentation Comment](https://assets.postman.com/postman-docs/doc-comments.png)
```
claireriley, line 83: Let's zoom in tighter on comments here as it's a bit hard to read. Also would consider making a realistic comment (doesn't have to be complicated, could just be something simple like "Visit https://developer.accuweather.com/ to find more information on this API.")

claireriley (changes requested):
This looks great! Few rewording suggestions.

We'll want to make sure all screenshots are jpg—I think you can configure that when you export in snagit. You can run them through https://imageoptim.com/mac then upload to aws.

For styling, we'll want to make sure smaller screenshots of things like menus don't get blown up to page width as they can be a bit annoying to scroll through. Use your best judgement based off of your local (I would spin up with npm run dev even if you're using a preview in your editor to get the full experience). [Here's an example](https://learning.postman.com/docs/designing-and-developing-your-api/monitoring-your-api/viewing-monitor-results/)—if you scroll down, most images are pretty wide so full width makes sense, but then the activity log is much smaller. To compare [here's an example](https://learning.postman.com/docs/administration/managing-your-team/) where I haven't fixed sizing yet, so you have to scroll through these huge vertical screenshots

claireriley (approved):
Sorry for the delay on this one @jerry-li-dev ! Looks great, let's get it published 😄

claireriley:
@jerry-li-dev make sure to delete your branch after merging! You should see a button below to

PR #2729, opened 2020-12-07
Requests doc updates for app V8

SueSmith:
Mostly pics..


claireriley (approved):
Thanks @SueSmith ! Looks great, we'll merge with the rest of the v8 docs 🍻

PR #2731, opened 2020-12-07
removed date from all blog URLs

tristandenyer:
I removed all the dates from the blog URLs (an old format) to reduce reliance on redirects.


ch264 (approved):
LGTM

PR #2766, opened 2020-12-22
More tweaks to Homepage

arlemi:
Added livestreams to the bottom of the homepage with a custom component to track outgoing links in GA. 
Did not include Office Hours for now as I thought It'd probably be too long, and we may want to add them somewhere else as they're a different type of content.


SueSmith (approved):
ran locally and had a wee look, lgtm!

loopDelicious (commented):
looked good on @SueSmith 's machine!

PR #2774, opened 2021-01-05
Added livestream to homepage

arlemi:
Following our discussion around adding more "dynamic" content to the Homepage.


SueSmith (approved):

SueSmith:
@arlemi lgtm! are we going to be updating it when events pass? since one is tomorrow 😅

PR #2785, opened 2021-01-15
V8 api workflow

arlemi:
Updated vocabulary and screenshots, I haven't updated the screenshots for creating the different elements as they are cropped enough that it looks the same on both UIs.


claireriley (approved):
Thanks @arlemi ! Looks great—made a few minor adjustments

PR #2802, opened 2021-01-21
Delete hipchat and add redirect to main page

jerry-li-dev:
- Deleted hipchat integrations page
- Redirect to main integrations page


jerry-li-dev:
HipChat integration is no longer supported. We need to remove this page and redirect to the main integrations page.

tristandenyer:
@jerry-li-dev There is a link left behind on `changelog-and-restoring-collections.md` to old hipchat page that needs to be removed. And there are 3 old redirects still pointing to old hipchat page that need to be updated to main page. Search for `"to": "/docs/integrations/available-integrations/hipchat/"` in redirects file.

tristandenyer (changes requested):
see comment

tristandenyer (approved):
LGTM

PR #2804, opened 2021-01-21
Specify data parameter used while scripting in visualizations is a JSON object

meenakshi-dhanani:
The data parameter used in pm.visualizer.set is a JSON object. Since the documentation mentions object, it could make the user feel that it is a Javascript object and they would try to pass other types of Javascript objects to it eg. functions which a JSON object does not accept. Hence the change.


claireriley (approved):
Looks great @meenakshi-dhanani ! I removed the extra space at the end as it was causing a linting error. Please merge and we'll get this pushed with the v8 release!

PR #2807, opened 2021-01-23
V8/setting up monitors

claireriley:


src/pages/docs/designing-and-developing-your-api/monitoring-your-api/setting-up-monitor.md
```diff
@@ -33,15 +33,15 @@ Postman Monitoring offers a number of configuration options when creating a moni
 
 * [Creating a monitor](#creating-a-monitor)
```
ArjunSingh-PM, line 34: I think we can change the order in which we present various ways to create a monitor. IMO, we should highlight the creation from the Sidebar and from collections first as these are primary flows. Thoughts?

ArjunSingh-PM (approved):
Looks good. I think we can re-look at the order in which we present various options to create a monitor. I would highlight creation from the sidebar and from a collection.

jerry-li-dev (approved):

src/pages/docs/designing-and-developing-your-api/monitoring-your-api/setting-up-monitor.md
claireriley: Makes sense. I've moved those up however left the create from Overview section first for now (since that section is the bulk of the screen when you open Postman or move between workspaces, I assume we want users to use it). We can see how this flows and continue to adjust

PR #2815, opened 2021-01-26
refactored homepage to reuse components and align sections

tristandenyer:
Refactored homepage to:
* reuse card components
* hero to better match latest hero layout on WWW, and improve breakpoint handling
* sections have standard responsive padding
* sections have standard max-width of content to help align content down left and right sides
* added in light blue background to break apart sections
* added in grid re-ordering for video so text is above video on small breakpoint
* refactored `LandingCard` background to support same-height cards
* refactored `LandingCard` image to be slightly bigger
* refactored heading tags for better SEO (H1 and H2)
* fixed orange links on hover to be black (like other pages) instead of blue
* fixed footer padding for small breakpoint to keep text from wrapping on iPhone X size.

localhost_8000_


ch264 (approved):
LGTM

PR #2816, opened 2021-01-27
fixed card hover state

tristandenyer:
Fixed Card (with circle image) hover state to show orange when user hovers anywhere on card.

Screen Shot 2021-01-26 at 4 53 23 PM


ch264 (approved):
LGTM

PR #2818, opened 2021-01-29
Update faq on monitor permissions

claireriley:


ArjunSingh-PM (approved):
Thank you @claireriley!

PR #2860, opened 2021-02-25
Dependencies update - Gatsby and Gatsby cli versions

brandonmcastillo:
Updating current Gatsby versions to be up-to-date.


tristandenyer (approved):
The ^ should be handling this anyway, but happy to update to latest version.

PR #2866, opened 2021-02-26
Adds alert box to contextual links

mclausaudio:
This pr adds an alert box to the right sidebar under the contextual links section if a developer does not have the .env.development file locally in repo.  This will usually be the case for external contributors, so this alert box lets them know that the repo is not broken if they are not seeing blog posts on certain pages.

Screen Shot 2021-02-26 at 2 44 15 PM


tristandenyer (approved):
LGTM

PR #2867, opened 2021-02-27
Updated echo links

### LINKED ISSUE #2858: Automatic Redirection of echo.getpostman.com to https://postman-echo.com
In this document, https://learning.postman.com/docs/sending-requests/certificates/ the <echo.getpostman.com> link mentioned is not automatically redirected to https://postman-echo.com, also the redirect page is not having the redirect URL shown as a link, but as a plain text causing confusion that the webpage is not active. 
**Screenshots attached** 
1. image
2. image

**Solutions Expected :**
1. Either we change the doc with correct link : https://postman-echo.com
2. Or we can make the redirection automatic :)

tristandenyer:
Fixes #2858 
Updated echo links to be inline code (they are referenced as code in this case), and added call to action link to Postman Echo Collection to keep connection.

### DIFF
--- a/src/pages/docs/sending-requests/certificates.md
+++ b/src/pages/docs/sending-requests/certificates.md
@@ -30,7 +30,7 @@ To manage your client certificates, click the wrench icon on the right side of t
 
 To add a new client certificate, click the **Add Certificate** link.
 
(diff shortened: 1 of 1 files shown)


brandonmcastillo (approved):
Pulled and ran. Links work. Approved.

PR #2871, opened 2021-03-03
refactored divs to use landmark tags for improved a11y

tristandenyer:
* wrapped left nav in <nav> landmark tag
* wrapped middle content area in <main> landmark tag
* wrapped right sidebar is <aside> landmark tag
* wrapped footer in <footer> landmark tag


brandonmcastillo (approved):
Changes look good. Approved.

PR #2883, opened 2021-03-09
datadog-update

jerry-li-dev:
Initial push for configuration steps update


shashankawasthi88 (commented):
@jerry-li-dev we can skip the section "Metrics and Tags forwarded to Datadog from Postman" for now as we are working on a new spec where we have renamed these metrics. We would add a new section corresponding to the renaming of the metric and how to upgrade to a newer version of Integration by next week

shashankawasthi88 (approved):
This looks good Jon

PR #2890, opened 2021-03-11
Add note about custom prefix for bearer token authorization

GooBall:
Should help people looking for a solution to postmanlabs/postman-app-support#4727


jkonrath-postman (approved):
Thanks for the help! This looks like a good addition - approved.

PR #2896, opened 2021-03-12
optimized hero img, added size attributes and responsive class

tristandenyer:
* optimized hero image on homepage (reducing it by 510kb)
* added in size attributes to img tag
* added responsive Bootstrap className


mclausaudio (approved):
LGTM

PR #2927, opened 2021-03-25
Dev --> Prod

tristandenyer:
* updates handling of preconnects for resources we rely on for first paint.
* adds crossOrigin attribute that was missing in preconnect


brandonmcastillo (approved):
LGTM

PR #2935, opened 2021-03-30
LC Bug: Update conditional for onClickOut function for mobile browsers

brandonmcastillo:
This fixes the on click bug we are coming across for mobile browsers. I added a conditional to check the search box output and to update state only if theres input.

 We are using the react-onclickout package so that when a user clicks outside of the search box, the menu will disappear. The bug only appears on mobile browsers and doesn't appear on web.


tristandenyer (approved):
Pulled, ran, LGTM

PR #2957, opened 2021-04-06
Update making-soap-requests.md

ksearle:
Added the SOAPAction header to request header instructions.  My SOAP service would not work without it.  I wrote it succinctly which makes it seem like this is always required, but I couldn't find this written anywhere.  Yours was the best instructions I found, so just added this.  If this is more of a conditional requirement based on the implementation of the SOAP service, I can rephrase to a `may need` instead of `will need`.  But either way, I think the information is valuable.


jkonrath-postman (approved):
Looks good

jkonrath-postman:
Thanks for the contribution @ksearle! I'll go ahead and merge this.

Note: I originally wanted to make this "may need" instead of "will need" but in digging around, it looks like this was made a requirement in SOAP 1.1 ages ago, so I'll stick with your wording.

PR #2970, opened 2021-04-08
Update Authorize using browser info

SabriH10:
Added information regarding authorizing using the browser


jkonrath-postman (approved):
Thanks for the contribution @SabriH10! 

I had to make a minor edit so the Markdown would pass the lint check, and I made a very tiny grammatical change, but otherwise this looks great - I'll go ahead and merge. Thanks again!

PR #2984, opened 2021-04-13
[MARKENG-92] remove lock dependency versions in package file

oswish:
**Ticket/Issue:**
MARKENG-92

**What are the changes?**
Branched from develop,  this locks dependency versions in the package file by removing carrots.

**Why make these changes?**
To keep updated minor versions from auto installing & potentially causing unexpected problems.

_*no visuals_


oswish:
It's not as simple as only removing the "^" character; doin that prompts a message that there are multiple versions of sharp and that we need to update all the dependencies that are dependent on it.

PR #2985, opened 2021-04-13
[MARKENG-92] remove carets in package file, except gatsby*

oswish:
**Ticket/Issue:**
MARKENG-92

**What are the changes?**
_Branched from `develop`,  this locks dependency versions in the package file by removing carrots, except for a range of Gatsby dependencies.

**Why make these changes?**
To keep updated minor versions from auto installing & potentially causing unexpected problems.

_*no visuals_


oswish:
Discussed w/ @tristandenyer & we've determined that we are going to abandon this work for now; _ain't broke, don't fix..._

PR #2993, opened 2021-04-15
remove dated react-responsive-embed package, replace with standard if…

mclausaudio:
…rame

Screen Shot 2021-04-14 at 3 09 40 PM


tristandenyer (approved):
LGTM

Issue #2884, opened 2021-03-09
Docs about Token Scanner feature is missing


iprithvitharun:
I couldn't find any docs related to the Token Scanner feature under Settings. 

However, I was able to find the following blogs.

https://blog.postman.com/postman-security-scans/
https://blog.postman.com/how-were-protecting-your-postman-api-keys-in-github/

These are all scattered and don't provide the complete picture of how Postman helps its customers secure their Collections.

jkonrath-postman:
This was fixed in https://github.com/postmanlabs/postman-docs/pull/3060 (The issue didn't auto-close because it was linked to the old PR.)

PR #3011, opened 2021-04-21
Security Squad - Token scanner and Audit report article #2884

iprithvitharun:
Related to issue #2884


jkonrath-postman (changes requested):
I've made some minor changes. @iprithvitharun I just sent you a slack, let's follow up offline.

PR #3018, opened 2021-04-22
updated start URL for manifest file

tristandenyer:
This updates the starting URL for the manifest file to be absolute instead of relative. (This is the URL people's homescreen app icons are set to—should they add LC to their homepscreen.)


jkonrath-postman (approved):
LGTM

PR #3019, opened 2021-04-26
Export environment

### LINKED ISSUE #2923: Environment Management and export documentation outdated/inaccurate
With the latest UI changes, there does not appear to be an environment export option and the "manage exports" (as well as manage collections) UI is entirely different.  Just as a user for ~two years, I was not able to find away to export an environment like I was able to in the past.

sowmyaraj92:
To fix #2923


jkonrath-postman (approved):
Looks good to me

PR #3020, opened 2021-04-26
Clarified that some listed options are actually conversion scripts

jkonrath-postman:


loopDelicious:
Hmm, I don't think cURL is an API specification format. And Swagger 1 and 2 are. It might be best to tag someone from Open Technologies to weigh in on this one @jkonrath-postman

sowmyaraj92 (approved):
LGTM! 🎉

PR #3044, opened 2021-05-03
remove URL and use root for manifest file

tristandenyer:
This reverts the start)url change back to using the root. See Chrome console WARN:

Screen Shot 2021-04-30 at 2 38 13 PM


brandonmcastillo (approved):
LGTM

PR #3048, opened 2021-05-05
Fixed build error for GATSBY_NEWRELIC_ENV env variable

brandonmcastillo:
We only want to use NR in production so I've updated the build and build:beta commands in our scripts. This also removes the log for a missing GATSBY_NEWRELIC_ENV environment during build.


tristandenyer (approved):
LGTM

PR #3067, opened 2021-05-17
Add summary, category level up video on public workspace sidebar

meenakshi-dhanani:
add-summary-level-up

Adding so that it's easier for users to find how to make their public workspace and its elements discoverable on the network.


sowmyaraj92 (approved):
Looks good! 🎉

PR #3068, opened 2021-05-17
MARKENG-230 - Update case study url, fixes redirect

mclausaudio:
Updated URLs to case studies.  Current URLs cause a redirect on our marketing page.  These new urls avoid redirects.


tristandenyer (approved):
LGTM

PR #3096, opened 2021-06-02
Changed `wrench` to `gear`

brianwhited:

PR #3097, opened 2021-06-02
Change wrench to gear

jkonrath-postman:
A continuation of #3096 - there were a few other places that the wrench icon was mentioned instead of the gear icon.


claireriley (approved):
LGTM!

PR #3100, opened 2021-06-02
Fix redirected links

jkonrath-postman:
Two links in the WebSocket doc don't have a trailing slash, so they hit a redirect.


claireriley (approved):
LGTM!

PR #2416, opened 2020-05-27
Add documentation for use with other Chromium flavours

l3nn4rt:

PR #3123, opened 2021-06-11
Use a flavor of chromium other than chrome with interceptor

jkonrath-postman:
This is based on PR #2416 

I couldn't cleanly merge the old PR because everything's moved, so I started a new one. I also streamlined this a bit and moved it into the troubleshooting section.


claireriley (approved):
LGTM!

PR #3154, opened 2021-06-22
Update configuring-scim.md

ArjunSingh-PM:
A user can generate multiple SCIM API keys. Creating a new SCIM API key does not make existing keys inactive. If a user regenerates an existing API ley, they have the option to delete the existing API key or to delete it after a period of 6 hours of key regeneration.


jkonrath-postman (approved):
LGTM

PR #3167, opened 2021-06-25
Update capturing-http-requests.md

jerry-li-dev:


jkonrath-postman (approved):
Changes look good, but can you mention (either here or offline) if this is associated with a Jira or issue? We've got a few that affect this file.

PR #3169, opened 2021-06-25
Add language to every fenced code block

jkonrath-postman:
Every fenced code block should have a language after the three backticks. It causes a warning on build otherwise.


tristandenyer (approved):
Pulled, ran, checked code output, LGTM

PR #3179, opened 2021-06-30
Update importing-and-exporting-data.md

jerry-li-dev:


jkonrath-postman (changes requested):
* Changes added look good.
* The Exporting Environments section needs some changes - it's not on the gear icon anymore. If you open an environment, the ... menu in the top right has an export. For some reason, on the Environments tab, each one doesn't have an export on the ... menu like collections. 
* For Exporting data dumps, they changed how this works. So the screenshot changes, but now instead of locally downloading JSON files, that button goes to a web page where you select collections or environments (or both) and request a data dump. Then after a wait, you can download a zipped export.

jkonrath-postman (approved):
LGTM

PR #3193, opened 2021-07-12
Update issue templates

jkonrath-postman:


jkonrath-postman:
@chrisdeso Adding you so you can take a look at the linkage to support in the yml file - let me know if you want anything else in there pointing to support.

.github/ISSUE_TEMPLATE/documentation-request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Documentation request
+about: Suggest ideas for new documentation
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
claireriley, line 11: I might change the example starter to be something more in line with the idea that they are probably coming directly from our docs and haven't found what they needed as I'd assume that's the primary flow for users filing issues here

"I'm always frustrated when [...]" -> "I'm having trouble understanding how to [...]" "I find Postman's docs to be lacking in detail for [...]" "After reviewing the docs, I'm still not sure how to [...]"

claireriley (approved):
Looks good! One suggestion.

PR #3212, opened 2021-07-19
Dev -> Prod

jkonrath-postman:
TW-75 Proxy response capturing and request grouping
MARKENG-378 bump node version in package to 12.13.0

PR #3215, opened 2021-07-19
Revert "Dev -> Prod"

jkonrath-postman:
Reverts postmanlabs/postman-docs#3212


ch264 (approved):
reverting doc is ok, node version needs to be updated

PR #3218, opened 2021-07-20
update on OKTA errors

yiantseng:
Instacart ran into some OKTA errors when provisioning due to the users already being sent an invite. Added some clarification here for future customers, let me know if I could have worded it better!


jkonrath-postman (dismissed):
Hi @yiantseng - thanks for this - can you make one little change and then I will approve?

Change the last line to "If this occurs, remove the pending invite from the Postman manage team settings, and Okta will then provision them.

jkonrath-postman (approved):
LGTM

PR #3222, opened 2021-07-22
[MARKENG-388] added gatsby package for link checker on build

tristandenyer:
**REVIEW, BUT HOLD ON MERGE.**

This package was added previously and tested out, cleaned up links, then removed.
Adding in now to test it out through build and deployment.
* added package
* added to gatsby config (no options)
* addressed and fixed one broken link (anchor tag). All links should be clean going into this.

To test locally:

1. pull down
2. npm i
3. npm run build (should run clean in terminal)

break a link URL or anchor tag, and rerun. (should see WARN and ERR in terminal, showcasing link you broke)

Screen Shot 2021-07-22 at 12 24 02 PM

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -77,6 +77,7 @@ module.exports = {
       options: {
         plugins: [
           'gatsby-remark-autolink-headers',
+          'gatsby-remark-check-links',
           'gatsby-remark-responsive-iframe',
           {
             resolve: 'gatsby-remark-prismjs',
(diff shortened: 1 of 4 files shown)


ch264 (approved):
ran, build, broke, did not build. gives a warning though warn Plugin gatsby-plugin-catch-links is not compatible with your gatsby version 2.32.13 - It requires
gatsby@^3.0.0-next.0

PR #3290, opened 2021-07-27
[MARKENG-206] Added Cookie Notification snapshot test

brandonmcastillo:
Added CookieAlert.spec.jsx file, similar to the header and footer.

I ran npm run test:update so the footer is up to date. 

This targets the third-party OneTrust cookie notification popup. If a user removes or updates this component, the build will break.


tristandenyer (approved):
LGTM

PR #3310, opened 2021-08-03
[MARKENG-449] installed typescript package to clear dep WARN

tristandenyer:
Screen Shot 2021-08-03 at 11 51 07 AM

* running `npm i` then `npm run build` no longer shows dep WARN in terminal
* * web app builds and runs as expected—no changes should be noticeable on front-end


ch264 (approved):
pulled, installed, build 

lgtm 🏄

PR #3356, opened 2021-08-06
[TW-143] Remove postman.postman.com links and use go.postman.co

jkonrath-postman:
There shouldn't be links to postman.postman.com - use go.postman.com instead.


tristandenyer (approved):
Tested all links on logged in state, LGTM

PR #3383, opened 2021-08-22
Added Cred Card Number and CVV support

HeberPcL:
In **postman-collection** repo I added support for **$randomCredCardNumber** & **$randomCredCardCVV**

https://github.com/postmanlabs/postman-collection/pull/1218

And here I'm adding an example of usage and result.


jkonrath-postman:
@HeberPcL Thanks for the change. It looks good, but let me wait until the change in postmanlabs/postman-collection#1218 goes through, and then I can add the docs.

PR #3389, opened 2021-08-25
escape '|' character

fanofxiaofeng:
escape '|' character otherwise the text will be influenced incorrectly

if '|' character occurs in a markdown table, (I guess) it will make the text display incorrectly

Let me show the text before and after this change
# Before change
image

# After change
image


jkonrath-postman (approved):
Looks good! I'll merge after it passes checks.

jkonrath-postman:
Hi @fanofxiaofeng thanks for catching this! I've merged your fix to the develop branch, and it will go out the next time we refresh the site. Thanks again!

PR #3400, opened 2021-08-27
added new landing card and blue link styles and removed old code

ch264:
Screen Shot 2021-08-27 at 3 22 56 PM

New fonts are in other branch


tristandenyer:
The Cards on smaller breakpoints don't sit correctly.
They all need to be in one row (you have 2 now).
And use `col-sm-6 col-lg-3` for the cards, like the WWW version.

Screen Shot 2021-08-27 at 4 54 19 PM

tristandenyer (approved):
Pulled, ran, resized, LGTM!

PR #3405, opened 2021-08-30
Markeng-561 swap bottom cards for new cardless component

ch264:
added new cardless cards at bottom and also added a grey line at top and bottom of component

Screen Shot 2021-08-30 at 6 45 51 AM


tristandenyer:
@ch264 Let's remove the `justify-content-center` className and allow then to left-align when / if they wrap.

Screen Shot 2021-08-30 at 9 40 01 AM

tristandenyer (approved):

PR #3413, opened 2021-09-01
Duplicated LandingCard component

arlemi:
Hey @ch264 :D we needed to enable tracking of outboud links in the LandingCard component so I duplicated the original one and created one specific for cards containing an outbound link. Let me know if this would work or if you'd rather handle it directly in the LandingCard component?


ch264 (approved):
LGTM 🙌

PR #3432, opened 2021-09-14
updated sidebar and code styles

tristandenyer:
* updated right sidebar styles to match Sandbox
* updated inline code styles


ch264 (approved):
LGTM

PR #3462, opened 2021-09-16
[MARKENG-686] [c] Added in Firefox backdrop-blur workaround

mclausaudio:
Added a work around for `backdrop-blur` incompatibility with firefox.

Screen Shot 2021-09-16 at 10 03 58 AM
Screen Shot 2021-09-16 at 10 34 39 AM


tristandenyer (approved):
pulled, ran, checked Chrome and Ff, LGTM. Good workaround

PR #3480, opened 2021-09-17
offsets headers so they clear sticky nav when deep linking to element…

mclausaudio:
CSS solution to offset deeplinks so section titles clear sticky top nav

Makes use of the "copy deeplink" icon that is visible on hover, which is generated from the gatsby-remark-autolink-headers plugin


ch264 (approved):
LGTM

PR #3489, opened 2021-09-20
Use jQuery to manually offset user window when clicking on and visiting deep links, clears sticky nav

mclausaudio:


tristandenyer (approved):
Pulled, ran, tested on all breakpoints, LGTM

PR #3497, opened 2021-09-22
Update matching-algorithm.md

jwbanning:
There is no mention of the matching algorithm using parameters (i.e querystring params) to match the corresponding examples. This is something that I feel should be included in the documentation.


jkonrath-postman:
I had to move this to another branch. See https://github.com/postmanlabs/postman-docs/pull/3499.

PR #3498, opened 2021-09-22
[MARKENG-719] updated LI bullets to accept p tag inside li

tristandenyer:
Markdown applies a <p> to contents in an LI that has an indented UL below it. Targeting the li > p left an odd space to the left, so had to revert how we do bullets for these.

BEFORE
Screen Shot 2021-09-22 at 10 07 21 AM

AFTE
Screen Shot 2021-09-22 at 10 07 03 AM
R


mclausaudio (approved):
LGTM

brandonmcastillo (approved):
LGTM

PR #3500, opened 2021-09-22
Downgrade Bootstrap and jQuery packages to match WWW

brandonmcastillo:
Removed ^ symbol to avoid upgraded packages.
I also removed body-scroll-lock package since it is not being used in the navbar.


mclausaudio (approved):
LGTM

PR #3501, opened 2021-09-22
Remove flipped home icon in global nav

brandonmcastillo:
Icon was causing issues on Safari. I removed the flipped icon css and updated the HTML to reflect the changes.


tristandenyer (approved):

ch264 (commented):
LGTM

PR #3507, opened 2021-09-23
Update screenshots

sowmyaraj92:
* Installation and settings


jkonrath-postman (approved):
I backed out the change from introduction and moved it to a different branch (we talked offline about it...) 

Otherwise LGTM

PR #3510, opened 2021-09-23
search and replace for spurious character --> regular spacebar space

tristandenyer:
- did a search and replace for the spurious character ( , or U+A0, nonbreaking space), and replaced with regular spacebar space.
- this is to see if this clears up build errors we are seeing.
- previous commit we saw this character broke an anchor link, so the hypothesis is that it is not being recognized as a true nonbreaking space on build
Screen Shot 2021-09-23 at 11 08 57 AM


ch264 (approved):
lgtm

jkonrath-postman (approved):
LGTM

PR #3511, opened 2021-09-23
update ✦ to use CSS code value

tristandenyer:
in the effort to remove all the pasted in symbols and non-alpha and non-numeric values, changing ✦ to use CSS code value in CSS


ch264 (approved):
LGTM ✦

PR #3514, opened 2021-09-23
[MARKENG-721] Codeblocks styles updates, remove orange side margin and change the margin size

pamelagilmour:


src/templates/doc.scss
```diff
@@ -98,6 +98,11 @@
       padding-bottom: 2px;
       text-decoration: none;
     }
+
+    .before:hover {
```
tristandenyer, line 102: @pamelagilmour did you mean `:before:hover`? Not sure we have className of `before`. And I think the ordering for this would be `:hover:before`.

tristandenyer (changes requested):

src/templates/doc.scss
pamelagilmour: Oh this was work from a different branch that got brought over. I will fix now.

PR #3515, opened 2021-09-23
Dev --> Prod

tristandenyer:
- search-and-replace for the spurious invisible character that caused issues with the anchor link (and may be causing left nav issues.)
- removed calling Google for fonts. Now self-hosting.
- removed preconnects for Google fonts.
- removed commented out requests in head
- fixed hover state of left nav 2nd level down parent item
- updated Jest package to clear medium level vuln in terminal


ch264 (approved):
lgtm

PR #3517, opened 2021-09-23
updated inline code block styles + line-height for <p>

tristandenyer:
- updated the inline <code> to match styles in Sandbox (see screesnhot)
- update <p> tag line-height to match styles in Sandbox

Screen Shot 2021-09-23 at 2 11 53 PM


brandonmcastillo (approved):

ch264 (approved):
lgtm

PR #3518, opened 2021-09-23
[MARKENG-723] removed hover state from linked images

tristandenyer:
Removes this hover state on linked images, while retaining the bottom-border on inline links.

Screen Shot 2021-09-23 at 3 21 57 PM


ch264 (approved):
LGTM

PR #3521, opened 2021-09-24
changed search result styling

ch264:
Screen Shot 2021-09-24 at 9 13 10 AM


tristandenyer:
@ch264 
* the color for the pagination should be #6b6b6b
* the hover background color should be #F2F2F2
* the color for "<-- Previous" should be the same color as the 2, 3, 4, ... #6b6b6b

tristandenyer (approved):

PR #3524, opened 2021-09-24
Dev --> Prod

tristandenyer:
- updated Degular font files
- added in Firefox specific styling to control the bulky fonts it was printing (looks more like Chrome and Safari now)
- localized Inter and IBM Plex, and added in font file support down to IE6 (for Inter and IBM Plex; Degular is only 'modern browsers')
- converted older CSS to use reusable font variables
- adjusted font-weight of top navbar to be more consistent
- updated colors to use reusable color variables
- updated Algolia search results pagination styling
-


jkonrath-postman (approved):
Confirmed that this is all you and no content.

brandonmcastillo (approved):
LGTM

PR #3556, opened 2021-10-02
Used `restore` instead of `recover`

### LINKED ISSUE #3402: Inconsistent terms in Restoring deleted collections
**Describe the bug**
* In Postman app it is used “Restore” but it is mentioned “Recovering deleted collections”. It will not be helpful for user when searching.
* Screenshot irrelevant. You could use a restore screenshot.
*Recover a deleted collection that is larger than 30 MB - Can be displayed in the top as it is common to the three points displayed.

**Pages affected**
https://learning.postman.com/docs/sending-requests/intro-to-collections/#recovering-deleted-collections

**Screenshots**
image

Let me know if I can solve this. Already I have made some progress.

soham4abc:
Signed-off-by: Soham Banerjee <sohambanerjee4abc@hotmail.com>

Fixes #3402 

Please do have a look @jkonrath-postman


jkonrath-postman:
Thanks @soham4abc for the PR, but I looked at this and we can't change the verbiage on this - see my comment on the linked issue.

PR #3658, opened 2021-10-29
[MARKENG-862][c] use the pmTech payload from pm-monitor for SDK

oswish:
**What are the changes?**
_Branched from `develop`_, this uses the pmTech payload from pm-monitor.

**Why make these changes?**
As we develop the pm-tech SDK, it's version number increments. Using `npm` for version management results in several PRs with one-digit changes. This eliminates that. It also, subsequently, resolves an issue where OSS developers might collide with the NPM_TOKEN environment variable.

_*no visuals (regressive)_


tristandenyer (approved):
Pulled, ran, LGTM

PR #3675, opened 2021-11-04
[MARKENG-957][c] eliminate potential race condition for load events with pm-tech

oswish:
**What are the changes?**
_Branched from `develop`_, this embeds code (at build time) to eliminate a possible race condition, seen w/ some clients (Newrelic), do to timing on loading the page versus loading & parsing the SDK.

**Why make these changes?**
The is the second step; first step was safe-gurading pmt-tech functions, to eliminate q potential race condition, seen (a small amount) for some browsers, on the NewRelic dashboad.

_*no visual (regressive)_

### DIFF
--- a/bff.js
+++ b/bff.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const sh = require('shelljs');
+const base64 = require('base-64');
 const crypto = require('crypto');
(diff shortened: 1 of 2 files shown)


brandonmcastillo (approved):
LGTM.

PR #3682, opened 2021-11-08
TW-247 Additional tokens for token scanner

jkonrath-postman:


src/pages/docs/api-security/token-scanner.md
```diff
@@ -46,18 +46,29 @@ The Token Scanner will scan a variety of tokens by default. You can also add you
...
 * Bearer Token
+* Clojars Deploy Token
+* Databricks Authentication Token
 * DSA Private Key
 * EC2 SSH Private Key
+* Firebase Cloud Messaging API Key
 * GitHub Personal Access Token
 * Google API Key
 * Google OAuth Token
+* Microsoft Outlook Team Webhook
```
shreyaskm123, line 60: We can make this as Webhook URL to be more consistent with Slack Webhook URL

PR #3693, opened 2021-11-11
[MARKENG-980][c] newer gatsby-cli is causing console error; this specifies an older version, but latest patch possible

oswish:
**What are the changes?**
_Branched from `develop`_, this specifies an isolated version (v2.12.69) of gatsby-cli in the package json file.

**Why make these changes?**
In recent weeks, there has been a console error, when doing npm run dev, tied to the micromark-extension-mdxjs package. It was determined to be a dependency brought in by gatsby-cli.

Before:
LC-before

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -135,134 +135,104 @@
       }
     },
     "@babel/code-frame": {
-      "version": "7.12.11",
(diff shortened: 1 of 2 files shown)


pamelagilmour (approved):
LGTM, package-lock.json is nuts aka very long.

PR #3699, opened 2021-11-12
Dev --> Prod

tristandenyer:
* update target blank rel's
* update Launch button toggle to add delay until it gets the value from localstorage


brandonmcastillo (approved):
LGTM

PR #3713, opened 2021-11-29
[MARKENG-1052][c] simplify process to consume pm-tech SDK

oswish:
**What are the changes?**
This simplifies the configuration for consuming the pm-tech SDK

**Why make these changes?**
There are several web apps that are consuming & configuring the SDK. This aims to simplify the process & make as consistent as we can.

_*no visuals (regressive)_


brandonmcastillo (approved):
LGTM

PR #3726, opened 2021-12-03
TW-120  Echo page refresh

jkonrath-postman:
* New screenshot
* Minor edits

### DIFF
--- a/src/pages/docs/developer/echo-api.md
+++ b/src/pages/docs/developer/echo-api.md
@@ -33,7 +33,7 @@ To try the Echo API, open a new request in Postman and enter the following path 
 postman-echo.com/get
 ```
 
-Make sure `GET` is selected in the method drop-down, then select __Send__. The Echo API will return a JSON object that contains details from the request.
+Select the `GET` method, then select __Send__. The Echo API will return a JSON object that contains details from the request.
 
 [![Postman Echo response](https://assets.postman.com/postman-docs/postman-echo-api-response-v9.jpg)](https://assets.postman.com/postman-docs/postman-echo-api-response-v9.jpg)
 
--- a/src/pages/docs/developer/echo-api.md
+++ b/src/pages/docs/developer/echo-api.md
@@ -18,15 +18,15 @@ warning: false
 
 ---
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/developer/echo-api.md
```diff
@@ -18,26 +18,26 @@ warning: false
```
jkonrath-postman, line 18: Add this line to the front matter at the top, after line 4:

`updated: 2021-12-06`

jkonrath-postman (commented):
One little change, to add the `updated` date at top.

(I can't approve this because I initially created it. @mark-anthony-postman can, though.)

src/pages/docs/developer/echo-api.md
```diff
@@ -18,26 +18,26 @@ warning: false
...
+To try the Echo API, open a new request in Postman and enter the following path in the URL field:
 
 ```http
 postman-echo.com/get
 ```
 
-Make sure `GET` is selected in the method drop-down, and click __Send__.
-
-[![Postman Echo response](https://assets.postman.com/postman-docs/postman-echo-api-response.jpg)](https://assets.postman.com/postman-docs/postman-echo-api-response.jpg)
+Make sure `GET` is selected in the method drop-down, then select __Send__. The Echo API will return a JSON object that contains details from the request.
```
mark-anthony-postman, line 35: https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/d/dropdown
Per Microsoft, "dropdown" (no hyphen) should only be used as an adjective, so you could say "in the method dropdown list". Or you could simply avoid talking about UI items as Microsoft recommends and say something like "Make sure GET is selected for the method, then select..."

mark-anthony-postman (approved):
Made one minor comment, otherwise LGTM! 🙂

src/pages/docs/developer/echo-api.md
HeatherStenson: Thanks @jkonrath-postman! Updated in ffa10d4

src/pages/docs/developer/echo-api.md
HeatherStenson: Thanks @mark-anthony-postman, good call! Updated in f130af8

PR #3731, opened 2021-12-06
removed all ^; upticked packages to what ^ was using; swapped out sass package

tristandenyer:
* removed all ^ from package.json
* up-ticked package versions to use version package-lock was using from ^
* swapped out deprecated node-sass package
* no functionality changes

This is all to:
* lock packages so that rogue / unplanned npm updates don't keep us from deploying
* give us better control over updates
* move away from depreciated packages


mclausaudio (approved):
LGTM

PR #3745, opened 2021-12-09
The Postman chrome extension is deprecated

mukeshbolly:
As per the chrome web store, the Postman extension for chrome is deprecated and recommend to use the native app.


jkonrath-postman:
Hi @mukeshbolly thanks for the PR. The Postman Chrome extension is deprecated, but it still works. I can't remove this section until it's fully unsupported.

PR #3756, opened 2021-12-16
Refactored Left Nav to reduce flicker and maintain structure when navigating through pages

brandonmcastillo:
- Created Class component for LeftNav to use state
- Broke out the sectionHandler() and renderTwoLevelList() functions outside of class component
- Why? Calling the functions inside the render causes the component to completely re-render when a user clicks on a link
- Updated test due to capitalization on L and l for leftNavLinks

New behavior:
- No flickering up and down when user clicks link
- Side nav maintains active link
- Navigating between sub menus is much cleaner 

For visual, compare current side nav versus this branch locally.

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -7,8 +7,8 @@ import './styles/config/print.css'
 import $ from 'jquery';
 import 'jquery.scrollto';
 
-/* eslint-disable import/prefer-default-export */
 export const onClientEntry = () => {
+  console.log(!window.location.hash)
   if (!window.location.hash) {
     window.scrollTo(0, 0);
   } else {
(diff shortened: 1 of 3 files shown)


mclausaudio (approved):
LGTM

tristandenyer (approved):
pulled, ran, clicked around, console is clear, LGTM!

jkonrath-postman (approved):
LGTM but @brandonmcastillo one thing I don't know how to test locally - does it fire events correctly for Google Analytics and pm-tech? I want to make sure that clicking on three different left nav items shows up as three different distinct pageviews in analytics. Can we confirm that somehow?

brandonmcastillo:
> LGTM but @brandonmcastillo one thing I don't know how to test locally - does it fire events correctly for Google Analytics and pm-tech? I want to make sure that clicking on three different left nav items shows up as three different distinct pageviews in analytics. Can we confirm that somehow?

Yes, click events will be sent. I'll just need to add data-click attributes to the links before I merge this in. Note - this will be a lot of data if we go this route.

PR #3758, opened 2021-12-16
[MARKENG-1080] Adds modal functionality to full width images in docs

brandonmcastillo:
This adds the clickable modal functionality to full width images on doc pages. We are not targeting every image due to responsive issues.

- The `useModal` function takes an argument `parsedHtml`. 
- Query all the images on the doc and loop over
- If image does not have a width declared in the  element, allow modal. 
- Assign a unique ID to every the image since modal relies on data-target and IDs matching.

Functionality: 
- Pointer will display if image is clickable
- Close modal by clicking outside of image or on X


jkonrath-postman (commented):
It worked for me, but it was a bit... odd. 

The X in the upper right doesn't look great, and when it's on narrower images, it is really far away and it's not entirely obvious that it's there.

Would it be possible to move the close X so it's closer to the edge of the image? Or could the user click anywhere on the image to close it?

image

jkonrath-postman:
It's still really hard to see how you close an image. I would intuitively think the close should be in the upper right corner. Also, because it's gray with a transparent background, it gets lost. 

I also would like it if clicking anywhere on the image closes it, but I get it if that's not an option.

image

jkonrath-postman (changes requested):
See comment

brandonmcastillo:
@jkonrath-postman @mark-anthony-postman
 I updated the modal. 
- Updated logic to target missing images
- Centered images
- Removed "x" in favor of click anywhere to close the image

Very few images inside the repo are extremely large in height. We'll need to resize these to accommodate the modal. Adding a fixed height or padding to images causes the smaller images to blur out.

@mark-anthony-postman Do you have an example of the inline icon image? I couldn't find it.

mark-anthony-postman:
> @mark-anthony-postman Do you have an example of the inline icon image? I couldn't find it.

@brandonmcastillo For some examples, check out this page--see steps 2 and 3 under "Documenting a collection":
/docs/publishing-your-api/documenting-your-api/#documenting-a-collection

brandonmcastillo:
@jkonrath-postman @tristandenyer I updated the PR. 

Our docs contain a substantial amount of small images and will display blurry on laptop and ultra-wide screens so to counter act this, I disabled this functionality.

Workaround - Images that do not have a declared width in the  tag inside the MD doc will be clickable. 

This goes hand-in-hand with Jonathans original idea in targeting full width images only.  My proposed solution of targeting all images was causing many responsive issues so I reverted back.

@mark-anthony-postman I updated the odd hovering from the icons.

tristandenyer:
* I too feel we need to improve the close / X, make it more noticeable. I don't see one—perhaps it is under the top nav?
* for images like `/docs/getting-started/installation-and-updates/#selecting-the-agent-for-requests` the bottom gets cut off since it is only setting the width. Perhaps the actual modal is a container that is 90% width and height, and the media is a background-image for that container with `contain` on it? Let the CSS do the math to ensure the entire image shows.
* When I am viewing the above image and resize the viewport, the image goes back down to the original size. I think the 90% width and height can be applied to all breakpoints. On very wide screens, the 90% height would keep it all in check to not get too wide.
* using `esc` key should close modal.
*

brandonmcastillo:
@tristandenyer 
Using 90% height would cause small images to stretch and become blurry. 

I added the X back and left it open for styling. Modal can use esc key to close modal.

jkonrath-postman:
* If the image is at the top of a page, it gets eaten by the top scroll bar and you can't see the X: 

image

* It looks slightly off when the images are left-justified. Would it look better if they are centered?

tristandenyer (approved):
LGTM

jkonrath-postman (approved):
Just tested again, and looks good!

PR #3792, opened 2022-01-03
[MARKENG-994] Update deeplinking scrolling issues on Doc pages

brandonmcastillo:
This fixes two issues when navigating through deeplinks.

- Fixes issue when user clicks on an anchor link and the title hides behind the navbar. 
- Fixes issue when user clicks on a Docs link outside of Postman and page does not scroll to hash location.


mclausaudio (approved):
LGTM

PR #3806, opened 2022-01-10
[MARKENG-1161] updated date in footer, and test snapshots

tristandenyer:
* updated date in footer
* updated test snapshots
* updated print.css (date is used there as well)


chienleow:
LGTM

chienleow (approved):

PR #3809, opened 2022-01-10
fix class to be className

tristandenyer:
* changed class to be className on Last modified component, console error in dev mode


mclausaudio (approved):
LGTM

PR #3812, opened 2022-01-11
Update edge case bug for "Last modified" date

brandonmcastillo:
http://localhost:8000/docs/designing-and-developing-your-api/the-api-workflow/
https://learning.postman.com/docs/designing-and-developing-your-api/the-api-workflow/

This adds a p tag to the text element and allows it to render below an image instead of the side.
Screen Shot 2022-01-11 at 2 48 53 PM


jkonrath-postman:
Another example: `https://learning.postman.com/docs/designing-and-developing-your-api/creating-an-api/`

jkonrath-postman (approved):
Works for me locally. I put a second example in comments, if you need to test that one too.

PR #3816, opened 2022-01-13
[MARKENG-1079] Previous and Next link functionality 

brandonmcastillo:
This adds the functionality to navigate between previous and next links relative to the section the user is on.

- Due to the triple nested left nav object, I mapped, looped and filtered by index to grab the matching path name and display the previous or next index if it exists
- Broke out PreviousAndNextLinks.jsx under /components/modules/
- Keeps doc.jsx clean 
- Allows LeftNavItems.jsx to scale if a new doc and navigation link is added
- Previous and Next links will stay in their position and not shift around the container

To test, run dev or build and go to /docs/. Scroll to the bottom and navigate between links.


jkonrath-postman (changes requested):
A couple of things:
* This only works within a folder. So if you get down to getting-started/troubleshooting-inapp/, it shows no next link, but the next topic would be in the next folder, sending-requests/requests/.
* Having this below the date and bigger makes it look disjointed. Should the date be below the prev/next?
* Having a **Next** heading is confusing in that we have several topics with a **Next steps** heading at the end, which is usually free-form links to other related sections.
* Maybe instead of Previous and Next headings, we have a single line that is like this?

`<< Creating a collection (a bunch of space here)  Creating a workspace >>` 

I'm not a designer, we probably need someone else to make this look right. I can ping you a few examples offline of maybe how we could go with it.

jkonrath-postman (approved):
LGTM. There's probably more font futzing that could be done with the size and placement of the links, but works for me.

tristandenyer (approved):
LGTM!

PR #3829, opened 2022-01-19
Updated details on the steps and added FAQ

yiantseng:
I added more details on the admin/user experience from our support checklist. I have also added an FAQ section from the support checklist that goes over each scenario of a captured user. I haven't had time to format everything properly, please review and update, as the current iteration is not as informative for the customer as the internal support checklist.


claireriley:
Moving these updates to #3845

PR #3847, opened 2022-01-27
[339] Resolve LC cookie notice scroll issue + clear errors on homepage

brandonmcastillo:
1. To resolve external a links from scrolling after click, I installed https://www.gatsbyjs.com/plugins/gatsby-plugin-catch-links/
This will need to be tested in beta since localhost does not automatically scroll. I suspect the upgrade to Gatsby V4 could  be the reason as our other platforms don't have this scroll issue.

2. Cleared up SVG errors on homepage
Screen Shot 2022-01-27 at 11 47 18 AM

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -35,6 +35,12 @@ module.exports = {
       },
     },
     {
+      resolve: `gatsby-plugin-catch-links`,
+      options: {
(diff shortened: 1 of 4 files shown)


tristandenyer (approved):
Pulled, ran, LGTM

PR #3851, opened 2022-01-28
markeng-202

ch264:
broke out dropdown from header into Dropdown component and implemented clickout with hooks instead of third party package


brandonmcastillo (approved):
Dropdown component works! LGTM.

PR #3852, opened 2022-01-28
[MARKENG-339] Remove scroll packages; not working in develop / production

brandonmcastillo:
Removed gatsby-plugin-catch-links package. Issue for scroll top for One Trust cookie still remainds.
The issue is that locally, the page does not scroll to the top when compared to beta which does scroll. I suspect this could be coming from an update in Gatsby V4 as the issue does not persist on WWW which is on V2.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -35,12 +35,6 @@ module.exports = {
       },
     },
     {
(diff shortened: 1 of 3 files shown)


ch264 (approved):
LGTM

PR #3854, opened 2022-02-01
gatsby-react-router-scroll added back to develop

brandonmcastillo:
Missing Gatsby package isn't in develop and causing develop build to break.


mclausaudio (approved):
lgtm

PR #3856, opened 2022-02-02
[MARKENG-1212] Improved navigation for "Previous" and "Next" links for Docs 

brandonmcastillo:
**Bug/Issue:** Previous and Next links are hitting a dead end when navigating between a second level folder above or below the section causing the UI to feel broken. Current logic does not support this case.

**Solution:** 
I updated the logic into two parts. The first part filters out the left nav data depending on window location. The second part checks which active  link the user is on then pushes items into an array based on index and availability [last item from previous section, current items in active section, first item in next section]
- Added React Hooks useState and useEffect. Cleaner imo.

**Expected functionality:**
- Adds the ability to navigate docs another level deeper using "Previous" or "Next" methods (mouseclick, arrow keys on web or swipe on mobile)
- Example: Designing and Developing your API <-> Mocking Data <-> Analyzing Reports

✅

Screen Shot 2022-02-01 at 7 17 32 PM

- Does not apply between sections 
- Ex: Getting Started <-> Sending Requests

 ❌

Screen Shot 2022-02-01 at 8 11 28 PM

**Other:**
- Updated comments
- Swipe and arrow key functionality still working

### DIFF
--- a/src/components/LeftNav/LeftNav.jsx
+++ b/src/components/LeftNav/LeftNav.jsx
@@ -8,10 +8,9 @@ const sectionHandler = (e) => {
   document.location.href = e.target.getAttribute('data-section');
 };
 
-const renderTwoLevelList = (item, runtime,) => {
+const renderTwoLevelList = (item, runtime) => {
   if (typeof document !== 'undefined') {
     const active = runtime ? document.location.pathname.match(item.parentSlug) : '';
-  
     return (
       <ul key={uuidv4()}>
         <li className="parent">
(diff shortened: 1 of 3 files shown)


ch264 (approved):
buil app and clcked around, LGTM

brandonmcastillo:
@jkonrath-postman @tristandenyer requesting review since you're both aware of the minor UX issues we've talked about. thanks!

jkonrath-postman (approved):
Works for me.
There's a separate side issue with the ordering of stuff in Administration, but that's our issue - I'll create another bug for that.

PR #3866, opened 2022-02-08
Markeng 1227 - Point gatsby-config to use env vars when generating  site map url

mclausaudio:
Right now, Beta's site map uses `postman.com` when it should be `postman-beta`.  This PR makes the config file grab  the URL from the env vars, which I added in GH secrets and set to beta and prod respectively./


brandonmcastillo (approved):

ch264 (commented):
LGTM

PR #3892, opened 2022-02-23
[TW-101] Refresh Requesting access page

mark-anthony-postman:
- Added details for requesting access to environments and mock servers
- Consolidated steps into a single common procedure rather than separate procedures for each element type
- Updated screenshots


HeatherStenson (approved):
Such a big improvement! Looks great to me 🍊

PR #3934, opened 2022-03-17
refactored Breadcrumbs to use <ol> making it more accessible and semantic

tristandenyer:
* refactored how we build the breadcrumb to use <ol>, which is preferred for a11y to show hierarchy
* applying `aria-hidden` to separators to hide from screenreaders
* added `title` to "Home" link to explore how this affects the SEO preview card on LinkedIn (which currently prints "Home" instead of "Learning Center")
Screen Shot 2022-03-17 at 10 48 53


jkonrath-postman (approved):
LGTM

PR #3946, opened 2022-03-22
feature/docs-test-execution

rtfmoz2:
Added order of test execution to postman docs, its not intuitive and the general assumption will be they run after.


jkonrath-postman:
Thanks for the contribution @rtfmoz2! Looks good, I'll go ahead and merge it. Thanks again!

PR #3947, opened 2022-03-22
added #icon to add CSS hook for inline icons

tristandenyer:
* Adding support support for appending inline icons with `#icon` as a CSS hook to help style them to be more inline with text.
* updated one page's icon to show as an example.
Screen Shot 2022-03-22 at 14 46 36


mclausaudio (approved):
LGTM

PR #3948, opened 2022-03-22
Updated print.css styles to target #icon and keep them from becoming block-level

tristandenyer:
* icons would blow up to block-level size on print 
* targeting #icon, we can now keep them inline and 18pt width

BEFORE:
Screen Shot 2022-03-22 at 16 22 19

AFTER:
Screen Shot 2022-03-22 at 16 25 03

### DIFF
--- a/styles/config/print.css
+++ b/styles/config/print.css
@@ -111,6 +111,12 @@
         height: auto;
         display: block;
     }
+    img[src$='#icon'] {
+        margin-bottom: 0;
+        display: inline;
+        width: 18pt;
+        height: auto;
+      }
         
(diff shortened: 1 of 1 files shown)


mark-anthony-postman (approved):
The change works great! Adding `#icon` to icon images keeps them from blowing up and stomping all over the page when printing 🦖

jkonrath-postman (approved):
LGTM!

Note to other reviewers: the only page you'll see this on right now is `http://localhost:8000/docs/administration/managing-your-team/user-groups/` (Until we change all other icons)

PR #3919, opened 2022-03-10
TW 64 Updated information about public collaboration

HeatherStenson:
- Moves /docs/collaborating-in-postman/public-workspaces/ into /using-workspaces/
- Updates internal links to account for ^
- Reformats content in public-workspaces.md
- Adds additional information per comments from @gitThingsDone in TW-64

PR #3956, opened 2022-03-28
Fix redirect vs page error

HeatherStenson:
Fixes accidental reintroduction of `public-workspaces.md` into the original path from https://github.com/postmanlabs/postman-docs/pull/3919.


jkonrath-postman (approved):
LGTM

PR #3965, opened 2022-03-30
Now using font files from Voyager; updated font-face and CSS

tristandenyer:
* now sourcing fonts from Voyager (same as WWW), which results in local cache being used instead of re-downloading fonts for those that traverse properties
* updated font-face and CSS to match WWW patterns
* deleted the old font files
* there should be zero shift in front-end typography


jkonrath-postman (approved):
Pulled and ran, didn't see any errors. LGTM.

PR #3978, opened 2022-04-05
Update interceptor.md

chnuessli:
from footer to header :)


jkonrath-postman:
Thanks for the PR @chnuessli but this actually just moved :) It's in the footer now. The old icon is still in the header, but it just opens a note to tell you to look below:

image

PR #3985, opened 2022-04-07
Replacing dabbing Postmanaut with Voyager sourced SVG to clear render error

tristandenyer:
* reported that the dabbing Postmanaut does not render for some visitors (I could not reproduce)
* we found this similar error happening in Chrome for some `data:image` embedded SVGs. Replacing them with Voyager sourced images mitigated the issue.
* replacing dabbing Postmanaut with Voyager sourced SVG to clear render error
* see right side bar of doc pages

Screen Shot 2022-04-07 at 09 31 30


mclausaudio (approved):
LGTM

PR #3989, opened 2022-04-07
revert YouTube video to not use `-nocookie` domain to clear CORS errors from Google

tristandenyer:
* YouTube video now uses youtube.com instead of youtube-nocookie.com
* this will clear the new CORS errors coming from Google Play

Screen Shot 2022-04-07 at 12 31 17


brandonmcastillo (approved):
LGTM

PR #3996, opened 2022-04-11
TW 435 Audit logs page

HeatherStenson:
- Updated content about audit logs
- Ran Vale (I can't resolve many of the suggestions about passive text because they are part of text in the UI)

I still need a screenshot of these updates and will work with Ronak to get one. 🤝

### DIFF
--- a/src/pages/docs/administration/audit-logs.md
+++ b/src/pages/docs/administration/audit-logs.md
@@ -19,7 +19,7 @@ Audit logs display events related to your team, users, and billing. Team admins 
 * When users were added to, removed from, or invited to your team
 * Which user performed a specific action and when they did so
 
-Audit logs are retained for 90 days for Professional plan customers, and 180 days for Enterprise plan customers. <!-- TODO: reword -->
+Postman provides audit logs for 90 days for users with Professional plans, and 180 days for users with Enterprise plans.
 
 > Audit logs are visible to all [team admins](/docs/collaborating-in-postman/roles-and-permissions/#team-roles) on [Postman Professional and Enterprise plans](https://www.postman.com/pricing), including super admins on Enterprise teams.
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/administration/audit-logs.md
```diff
@@ -14,36 +14,52 @@ contextual_links:
...
 
-You can view audit logs in Postman by selecting **Team** in the upper-right corner, then **Audit Logs**.
+You can view audit logs in Postman by selecting **Team** in the header, then selecting **Audit Logs**.
 
 <img alt="Team menu with audit logs selected" src="https://assets.postman.com/postman-docs/menu-select-audit-logs-v9.13.0.jpg" width="300px"/>
 
-Audit logs allow you to view team actions by user, event name, event description, and date.
+Audit logs provide information about actions taken by members of your team. Each event includes the user, the user's email address, the event name and a description of the event, the user's IP address, and the date and time of the action, including the user's timezone.
+
+Use the filters to filter the actions by date range, event type, and user.
```
claireriley, line 44: Maybe something like this to avoid using filter twice
```suggestion
You can use filters to view actions by date range, event type, and user.
```

claireriley (approved):
Looks good! Aside from screenshot, one minor suggestion.

src/pages/docs/administration/audit-logs.md
```diff
@@ -106,7 +122,7 @@ Postman logs events related to your [team](#team), [users](#users), and [billing
 | Sent Team Invite  | An invitation was sent to a user to join your team.  |
 | Approved Team Invite | An admin approved an email invitation. |
 | Rejected Team Invite  | An admin rejected an email invitation. |
-| Cancelled Team Invite   | An invitation for a user was cancelled.   |
+| Canceled Team Invite   | An invitation for a user was canceled.   |
```
jkonrath-postman, line 125: For all three of these - I am not sure we can switch from "cancelled" to "canceled" because it's spelled wrong in the UI.

jkonrath-postman (approved):
One comment, but otherwise looks good to me!

src/pages/docs/administration/audit-logs.md
HeatherStenson: Good call! This is much better 😁

src/pages/docs/administration/audit-logs.md
HeatherStenson: Oh, right! Duh 😁 🤦  I'll revert.

PR #4015, opened 2022-04-25
Dev --> Prod Apr 25, 2022

tristandenyer:
- updated Websocket doc
- reverted fonts to now use locally instead of Voyager


brandonmcastillo (approved):
LGTM

PR #4031, opened 2022-05-04
remove contextual_links - right nav bar - from Algolia Index

ch264:


tristandenyer (approved):

jkonrath-postman (approved):
I don't have a way to test, but that is what we want removed from the query, so LGTM.

PR #4046, opened 2022-05-10
Updated the Inter font to latest files; cleaned up font-face declarations to improve performance

tristandenyer:
* added new Inter font files
* converted font-weight specific font names back to using just `Inter`
* converted variables back to using `$Inter`
* updated `font-feature-settings` to match Product's CSS.
* updated `display` to use `block` to see if this improves performance and reduces FOUT
* reduced IBM font to use only woff2 and woff
* there should be zero changes on front end
* this should improve performance and reduce FOUT on page load


brandonmcastillo (approved):
Screen Shot 2022-05-10 at 4 05 27 PM
Screen Shot 2022-05-10 at 4 05 27 PM
 LGTM!

ch264 (commented):
LGTM 🍻

PR #4049, opened 2022-05-11
Query frontmattter md char length

ch264:
display console error if frontmatter length exceeds 9999 bytes on build


brandonmcastillo (approved):
LGTM

PR #4051, opened 2022-05-11
TW 450 Refresh formatting for Microsoft AD FS SSO page

HeatherStenson:
Afaik the actual instructions haven't changed here, so this was just to update formatting (it was kind of all over the place). I also ran Vale.


mark-anthony-postman (approved):
LGTM! 🌈

PR #4080, opened 2022-05-31
[TW-516] Change upgrade to update

claireriley:
- Changes "upgrade" to "update" (defines upgrade as a change in plan type, update as a change in app version) 
- Fixes all affected links
- Minor vale and readability improvements


HeatherStenson (approved):
Looks great! I tried it locally and the redirect worked as expected.

PR #4087, opened 2022-06-02
TW-519 Cycle 4 Vale additions

jkonrath-postman:


vale-styles/Postman/Avoid.yml
```diff
@@ -0,0 +1,26 @@
...
+  - app(?:lication)? file
+  - backbone
+  - backend
+  - contiguous selection
+  - you will see
+  - UI
+  - soon
+#  - new
+  - etc
+  - eg
```
HeatherStenson, line 19: Can we add "ie/i.e." as well? I found two of those in the LC yesterday 😅 

Also does there need to be another entry (or some regex magic) for "e.g." and "i.e." to catch differently-punctuated instances?

mark-anthony-postman (approved):
LGTM! 🙂

vale-styles/Postman/Avoid.yml
jkonrath-postman: I tried `ie` and it generated a million false positives.
Also this rule seems to have problems with tokens that have periods in them (see also Socket.io errors.) I'll add a note to look into this, though.

PR #4093, opened 2022-06-06
[TW-520] Doc Azure DevOps Git integration requires third-party access

mark-anthony-postman:


src/pages/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo.md
```diff
@@ -61,27 +61,35 @@ A typical workflow for API-first development with an external Git repo:
 
 You can connect an API to your remote Git-based repository, at the API level. This enables you to continuously synchronize changes between the repository and Postman. Versions and release tags are then synchronized between Postman and your Git repo.
 
-> The user account used for authentication requires full access to repositories. You may want to create a service account with limited permissions specifically for this integration.
+Keep in mind the following when connecting to a repository:
```
HeatherStenson, line 64: This is probably a style guide discussion, but maybe "consider" instead of "keep in mind" on the basis that it's simpler/less colloquial? (Or maybe "remember"?) 

```suggestion
Consider the following when connecting to a repository:
```

HeatherStenson (approved):
Looks good! 🍊

src/pages/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo.md
mark-anthony-postman: Personally, I think "keep in mind" sounds friendlier, and in doing a quick search I found many examples of its usage in documentation from Apple, Google, and Microsoft. While I agree we want our styles to reflect simpler language, I also think we want to avoid our language getting too stuffy or formal sounding.  I'm going to leave it here for now, but I added it to the backlog for discussion. Thanks for bringing it up! 😄

PR #4095, opened 2022-06-08
Updated all Gatsby packages to latest

tristandenyer:
* updated Gatsby framework package to latest
* updated supporting gatsby packages (those with `gatsby` in name) to latest
* no changes to front end should be seen
* 4.15 brings with it the Scripts component API—while I am not using it directly, we need to validate it did not disrupt the pmTech script and others.


tristandenyer:
Closing in favor of other PR to fix

PR #4103, opened 2022-06-15
TW-522 More Vale changes

jkonrath-postman:


HeatherStenson (approved):
Looks good! It'll be interesting to see whether/how much making the suggested improvements impact an individual page's Flesch-Kincaid score.

PR #4115, opened 2022-06-28
[MARKENG-1550] Clear up object-object display for slow connections

brandonmcastillo:
This clears up the [object object] text displaying when viewing the individual doc pages. Usually seen on slower connections.


ch264 (approved):
pulled, build, LGTM 🍻

PR #4123, opened 2022-06-30
update packages to latest and remove carets to lock down versions

tristandenyer:
* 3 packages had carets
* periodically this would uptick and cause a new package-lock file to generate
* we have a policy of locking packages to a version (no caret), and manually update
* no new package-lock file needed


ch264 (approved):
LGTM

PR #4161, opened 2022-07-15
[TW-580] Clarified that the update feature is only available on desktop

MatthewPinsker-postman:
Where we mention the update feature, I clarified that this feature is only available on the Postman Desktop App. In the Installing and updating topic, I updated mentions of the Postman Desktop App to follow the same format. Also, I fixed 2 Vale suggestions.


jkonrath-postman (approved):
Looks great!

HeatherStenson (approved):
🍊 ✨

PR #4178, opened 2022-07-26
[markeng-1509-header] update header component with styled components

OlliesWorld:
- updated header component with styled components. 
- deleted scss file 
- created styles.jsx file as it is a long file

### DIFF
--- a/src/components/Header/HeaderStyles.jsx
+++ b/src/components/Header/HeaderStyles.jsx
@@ -2,9 +2,10 @@ import styled from 'styled-components';
 
 export const HeaderWrapper = styled.div`
 .navbar-v6 {
-  background-color: ${(props) => props.theme.colors.grey_00} !important;;
+  background-color: ${(props) => props.theme.colors.grey_00} !important;
   margin: 0;
   font-family: ${(props) => props.theme.fonts.Inter};
+  font-weight: 400;
 
   &.nav-primary {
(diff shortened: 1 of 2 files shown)


src/components/Header/HeaderStyles.jsx
```diff
@@ -1,17 +1,18 @@
-@import '../Shared/_variables.scss';
+import styled from 'styled-components';
 
+export const HeaderWrapper = styled.div`
 .navbar-v6 {
-  background-color: rgba(255, 255, 255, 0.32) !important;
+  background-color: ${(props) => props.theme.colors.grey_00} !important;;
```
tristandenyer, line 5: @OlliesWorld you have an extra `;` here

src/components/Header/HeaderStyles.jsx
```diff
@@ -1,17 +1,18 @@
...
   margin: 0;
-  font-family: $Inter;
+  font-family: ${(props) => props.theme.fonts.Inter};
```
tristandenyer, line 7: wherever we use `font-family: ${(props) => props.theme.fonts.Inter};`, we need to also declare its font-weight as 400 or 600.

tristandenyer (changes requested):

OlliesWorld:
I added the font weights, fix dropdown padding and changed the button styles name to a more clear description.

PR #4181, opened 2022-07-27
[markeng-1509-footer][c] update footer to use styled components

OlliesWorld:
- updated footer file to use styled components
- deleted scss file
- added theme to testing


tristandenyer (approved):

ch264 (changes requested):
There are old commits in here. Let's investigate this before merging

PR #4185, opened 2022-07-28
Develop --> Production

tristandenyer:
Release notes:
- Fixed footer runtime bug where it collapsed down to one column after pageload—only affected homepage
- updated `$borderRadius` scss variables to use 10px grid, be more inline with WWW
- 83 Docs updated
- Number of right sidebar links updated


claireriley (approved):
LGTM! ⭐

PR #4194, opened 2022-08-01
Add Comment and SingleMainWindow to Linux Desktop File

amayer5125:
Add a `Comment` which is used as a tooltip in some desktop environments and `SingleMainWindow` which signals that only one window can be opened to the Linux Desktop file example.

### DIFF
--- a/src/pages/docs/getting-started/installation-and-updates.md
+++ b/src/pages/docs/getting-started/installation-and-updates.md
@@ -97,11 +97,13 @@ Enter the following content in the file, replacing `/path/to/Downloads` with the
 [Desktop Entry]
 Encoding=UTF-8
 Name=Postman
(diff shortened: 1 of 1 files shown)


jkonrath-postman:
Hi @amayer5125, thanks for the contribution. 
I don't think this is something we want to do, given a recent change in window/tab behavior in Postman.
We've started rolling out a new feature (which is opt-out) where tabs are not shown inside Postman for individual items, and browser tabs (or windows) are used instead.
I haven't tried this in Linux yet, but I could definitely see a situation where a person would want multiple windows based on that. 
I'll ask the team, but I'm not going to pick this up. Thanks for the contribution though, and feel free to make more in the future or let us know if there are any other questions.

PR #4198, opened 2022-08-02
[markeng-1509] update footer to styled component

OlliesWorld:
- deleted .scss file
- updated footer.jsx to use styled components
- updated spec file to use theme


brandonmcastillo (commented):
LGTM

PR #4223, opened 2022-08-13
workflows(actions/checkout version) bump all workflow actions/checkout versions to v3

ArjunSharda:


jkonrath-postman:
Hi @ArjunSharda - thanks for the contribution, but we looked into this and can't change this right now:

* requires [actions runner 2.285](https://github.com/actions/runner/releases/tag/v2.285.0)
* Only a single commit is fetched by default, the action needs to be rewritten to support all commits
* is updated to the node 16 runtime (we are on node 14 -> would need to test the implications

We can look into making the changes after our next major release, but for now, I'm going to close without merge.

PR #4227, opened 2022-08-18
updated pagination to use latest linkArrow styling

tristandenyer:
* updated the pagination to use latest link Arrow styling
Screen Shot 2022-08-18 at 12 54 54


ch264 (approved):
pulled, ran, lgtm

PR #4232, opened 2022-08-22
TW-531 MFA docs

claireriley:
Includes:
* Setting up 2FA
* Turning off 2FA
* Minor adjustments to sign in flow

Note: We're avoiding the term "disable" in docs so that is only used in button labels (see "Inclusive Language" in TW Confluence)


src/pages/docs/getting-started/postman-account.md
```diff
@@ -105,14 +109,46 @@ If you're on an Enterprise plan, you must contact your Team Admins to update the
...
 
 If you aren't signed in to your Postman account, you can recover your username or reset your password from the [**Sign In**](https://identity.getpostman.com/login) page. Select __Trouble signing in?__ Enter your account email, select the __I'm not a robot__ box, and select __Submit__. Postman will send you an email with instructions for recovering access to your account.
 
 > If you still have trouble logging into your Postman account, contact [Postman support](https://support.postman.com/hc/en-us).
 
 If you're on an Enterprise plan, you must contact your Team Admins to update the password associated with your Postman account.
 
+#### Setting up two-factor authentication
+
+If you don't sign into Postman through SSO or Google authentication, you can set up two-factor authentication for your account:
```
ArjunSingh-PM, line 122: IMO, we should reword this to specify when you can setup 2FA rather than when you can not. 

"If you sign in to Postman using a username/email and password, you can set up two-factor authentication for your account:"

CC @Suhas-Gaikwad

ArjunSingh-PM (approved):
Thanks, Claire. the doc changes look good. I have added a minor comment.

petewitcher (approved):
LGTM

src/pages/docs/getting-started/postman-account.md
```diff
@@ -105,14 +109,46 @@ If you're on an Enterprise plan, you must contact your Team Admins to update the
...
 
 If you aren't signed in to your Postman account, you can recover your username or reset your password from the [**Sign In**](https://identity.getpostman.com/login) page. Select __Trouble signing in?__ Enter your account email, select the __I'm not a robot__ box, and select __Submit__. Postman will send you an email with instructions for recovering access to your account.
 
 > If you still have trouble logging into your Postman account, contact [Postman support](https://support.postman.com/hc/en-us).
 
 If you're on an Enterprise plan, you must contact your Team Admins to update the password associated with your Postman account.
 
+#### Setting up two-factor authentication
+
+If you sign into Postman using a username/email and password, you can set up two-factor authentication for your account:
```
MatthewPinsker-postman, line 122: Since we use "2FA" to abbreviate "two-factor authentication", what do you think about defining this abbreviation next to the first occurrence of "two-factor authentication"?

For example: `If you've set up [two-factor authentication](#setting-up-two-factor-authentication) (2FA), Postman will...`

src/pages/docs/getting-started/postman-account.md
```diff
@@ -105,14 +109,46 @@ If you're on an Enterprise plan, you must contact your Team Admins to update the
...
+
+1. Open your [account settings](https://go.postman.co/settings/me/account/) page by selecting your avatar in the upper-right corner > **Settings** > **Account**.
+1. Select **Enable 2FA**.
+
+    > Enabling 2FA will revoke all of your [active Postman sessions](/docs/getting-started/postman-account/#managing-your-active-sessions).
+
+1. Install an authenticator app on your phone if you don't have one available, and have your password on hand.
```
MatthewPinsker-postman, line 129: What do you think about using something like "available" instead of "on hand"? My concern is that "on hand" is a phrase that might not be clear to people reading in a second language.

src/pages/docs/getting-started/postman-account.md
```diff
@@ -105,14 +109,46 @@ If you're on an Enterprise plan, you must contact your Team Admins to update the
...
+1. Select **Continue**.
+1. Enter your password and select **Verify**.
+1. Link your authenticator app by scanning the QR code, then enter the 6 digit authentication code generated by your app and select **Next**.
```
MatthewPinsker-postman, line 132: It's my understanding that scanning a QR code is not the only way to link an authenticator app. I believe you can also enter an activation code generated from the application you are linking to. This might depend on the authenticator app you are using.

To make this step more generic, what do you think about removing the following: `by scanning the QR code`?

MatthewPinsker-postman (approved):
Looks good to me! I added a few inline questions. Thanks!

HeatherStenson (approved):
Looks great! +1 on Matthew's notes in the "Setting up" section 🍊

PR #4253, opened 2022-08-31
Fix naming convention in searchStyles and searchPreview to fix Beta build

brandonmcastillo:
WARN is displaying when running `npm run dev` and is breaking merges into BETA.


ch264 (approved):
LGTM

PR #4255, opened 2022-09-01
[MARKENG-1635B] update 2nd nav bg color 

OlliesWorld:
changed the bg color to white so page doesn't display under sticky nav
- MARKENG-1742 modal close btn border radius to new px grid


ch264 (approved):
LGTM

PR #4281, opened 2022-09-19
[TW-686] Added note about automatic major updates

MatthewPinsker-postman:
In the Downloading Postman v9 section, noted that the app will be auto updated to the latest major version if auto major updates is enabled.

### DIFF
--- a/src/pages/docs/administration/updating.md
+++ b/src/pages/docs/administration/updating.md
@@ -100,6 +100,8 @@ If you'd like to downgrade to Postman v9 or below to continue working with a tea
 
 Use the following links to download the latest Postman v9 version for your platform.
 
+> If the Postman desktop app has automatic major updates enabled, the app will be automatically updated to the latest major version.
+
 * [Download Postman v9 for macOS (Intel chip) 64-bit](https://go.pstmn.io/dl-osx64-v9-latest)
 * [Download Postman v9 for macOS (Apple chip) 64-bit](https://go.pstmn.io/dl-osxarm64-v9-latest)
 * [Download Postman v9 for Windows 64-bit](https://go.pstmn.io/dl-win64-v9-latest)


src/pages/docs/administration/updating.md
```diff
@@ -100,6 +100,8 @@ If you'd like to downgrade to Postman v9 or below to continue working with a tea
 
 Use the following links to download the latest Postman v9 version for your platform.
 
+> If the Postman desktop app has automatic major updates enabled, the app will be automatically updated to the latest major version.
```
jonycheung, line 103: I think it may be the confusing since "automatic major updates" toggle is not visible for most users. Can we invert it so that it states that the Postman app automatically update always unless you have the automatic major updates toggle enabled for your team.

MatthewPinsker-postman: Thanks, @jonycheung. I see what you're saying. **How about this?**

"The Postman desktop app will update to the latest major version. If you turn off automatic major updates, the app will continue using v9."

**Question:** To clarify, the automatic major updates toggle is NOT specific to a Postman plan, correct? I'm asking because it would be extra clear if we could update the second sentence to something like: "If you turn off automatic major updates, the app will continue using v9 ([Enterprise teams only](https://www.postman.com/pricing/))."

jonycheung: > To clarify, the automatic major updates toggle is NOT specific to a Postman plan, correct?

No, this toggle is not specific to Postman plan. I think what you have looks good. We don’t want to be specific as the rules for showing the toggle changes

MatthewPinsker-postman: Thanks @jonycheung. I updated the note, and it is ready for final review.

jkonrath-postman (approved):
LGTM

PR #4284, opened 2022-09-19
[MARKENG-1550] Disable clickable modal due to page load speed

brandonmcastillo:
Pages no longer display object object text but doesn't load all together. This disables the modified HTML and clickable modal images until we find another solution.


ch264 (approved):
pulled, ran, build, image does not open in modal LGTM

brandonmcastillo:
@jkonrath-postman I ran into an issue where the doc pages do not load in unison on slower connections. This PR will disable the clickable modal images but I'd like some feedback since it's a functional doc change. I will be looking into alternative methods (_build instead of runtime_) to render the modal images.

PR #4294, opened 2022-09-23
Update interceptor.md to include links for manual install

mscottnelson:
Link for manual download / install of interceptor when extension fails to provide one. It was difficult to find this information, since my workflow did not proceed as described in the installation instructions. This should save other users some time when they run into the same issue.


jkonrath-postman (approved):

jkonrath-postman:
Hi @mscottnelson thanks for the contribution! I've merged this, and it'll be in our next production update. Thanks again!

PR #4306, opened 2022-09-28
[MARKENG-1831][c] update bff build scripts to use PM_TECH_RT to avoid potential rate limiting

oswish:
**What are the changes?**
_Branched from `develop`_, this updates bff build scripts to use PM_TECH_RT.

**Why make these changes?**
Doing so makes it so we avoid potential API rate limiting.

_*no-visuals_


ch264 (approved):
LGTM

PR #4355, opened 2022-10-24
Removal of repetition

w4dd325:
For the description 'A random boolean value' I removed the brackets containing (true/false) as this example exists in the 'Example' column. Also removed the second set of true and false in the examples column.

For the description 'A random bank account name' I removed the brackets containing (savings account, checking account) as there are already examples in the 'Example' column.

For the description 'A random transaction type' I removed the brackets containing (invoice, payment, deposit) as this example exists in the 'Example' column.

For the description 'A random company suffix' I removed the brackets containing (Inc, LLC, Group) as this example exists in the 'Example' column.

Changed '3 random paragraphs of lorem ipsum text' to be within the same cell and added <p> and <br /> tags to display the paragraphs as they appear when returned inside Postman.

Changed '1 to 5 random lines of lorem ipsum' to use <br /> tag instead of /n (which is displayed in plain text on the learning center page)

Formatted the markdown to make it easier to read.


jkonrath-postman (approved):
Looks good to me

jkonrath-postman:
Thanks @w4dd325! I've gone ahead and merged this. It will be in our next push to production.

PR #4403, opened 2022-11-10
Added a reference for Postman CLI and Newman

w4dd325:
This section says that the setNextRequest function "only works in Collection Runner", but this is slightly misleading as it also works with the command line too. 

I have altered the following section to include a reference to Postman CLI and Newman.

### setNextRequest() only works when you run an entire collection

The `postman.setNextRequest()` function has no effect when you run a request using **Send** and is only used when you run a collection using the Collection Runner, Postman CLI, or Newman.


jkonrath-postman (approved):
Looks good to me

jkonrath-postman:
Thanks @w4dd325 - I merged the changes and they will be in our next production update.

PR #4411, opened 2022-11-15
[TW-764] Clarified how to trigger the v10 upgrade prompt

MatthewPinsker-postman:
Also, changed occurrences of 'update' to 'upgrade' because this is a one-time action on an API. The header still uses 'update' because we link to this section from the product, and changing the header will change the URL.


mark-anthony-postman (approved):
LGTM! 😀

PR #4429, opened 2022-11-21
Add contextual link to updated Level Up

mjgutermuth:
Update link to point to the new video that deprecates the old video


claireriley (approved):
Thanks @mjgutermuth! We'll include this in our next push to prod 🙌

PR #4432, opened 2022-11-23
TW-759 Sentence case in leftnav and titles

jkonrath-postman:


src/pages/docs/postman-flows/core-concepts/blocks.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "About Blocks"
+title: "About blocks"
```
MatthewPinsker-postman, line 2: In the "Contents" section, should we add `## Contents` before the bullet points? Also, should we format each bullet point as sentence case, instead of title case?

MatthewPinsker-postman (approved):
Looks good!

PR #4474, opened 2022-12-19
TW 821 Clarify workplace roles availability

HeatherStenson:
Question for the group - Since the availability for these workspace roles is on Basic, Professional, and Enterprise plans and that's pretty lengthy for a list entry, I used "Paid plans only" as shorthand for that, with a link to the [pricing page](https://www.postman.com/pricing/). In your opinions, is that clear enough? Or should I use "Basic, Professional, and Enterprise plans only" instead? 🤔

### DIFF
--- a/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
+++ b/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
@@ -1,7 +1,7 @@
 ---
 title: "Defining roles"
 order: 72
-updated: 2022-05-11
+updated: 2022-12-19
 page_id: "roles_and_permissions"
 warning: false
 contextual_links:
--- a/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
+++ b/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
@@ -103,9 +103,9 @@ To learn how to manage team roles in Postman, see [Managing your team](/docs/adm
 
 You can [assign](/docs/collaborating-in-postman/using-workspaces/managing-workspaces/#managing-workspace-roles) three role types in Postman workspaces: **Admin**, **Editor**, and **Viewer**. Partner Workspaces offer an additional role type: **Partner Lead**.
 
-* **Admin** - Can manage workspace resources and settings
-* **Editor** - Can create and edit workspace resources
-* **Viewer** - Can view, fork, and export workspace resources
+* **Admin** - Can manage workspace resources and settings.
+* **Editor** ([Paid plans only](https://www.postman.com/pricing)) - Can create and edit workspace resources.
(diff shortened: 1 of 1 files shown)


src/pages/docs/collaborating-in-postman/roles-and-permissions.md
```diff
@@ -103,9 +103,9 @@ To learn how to manage team roles in Postman, see [Managing your team](/docs/adm
 
 You can [assign](/docs/collaborating-in-postman/using-workspaces/managing-workspaces/#managing-workspace-roles) three role types in Postman workspaces: **Admin**, **Editor**, and **Viewer**. Partner Workspaces offer an additional role type: **Partner Lead**.
 
-* **Admin** - Can manage workspace resources and settings
-* **Editor** - Can create and edit workspace resources
-* **Viewer** - Can view, fork, and export workspace resources
+* **Admin** - Can manage workspace resources and settings.
+* **Editor** ([Paid plans only](https://www.postman.com/pricing)) - Can create and edit workspace resources.
+* **Viewer** ([Paid plans only](https://www.postman.com/pricing)) - Can view, fork, and export workspace resources.
```
MatthewPinsker-postman, line 108: I'm leaning towards writing out "Basic, Professional, and Enterprise plans only", instead of "Paid plans only".

My thinking is that we note whether something only applies to one or two plan types on the same page. As a user, seeing (for example) "Enterprise only" and "Paid plans only" on the same page (or in the same section) forces me to recall which plans are paid plans. I might be more inclined to select "Paid plans only" to be reminded which plans fall in this category. I also think it might be easier to scan the content if the plan types are written out.

However, I see some occurrences of "paid plans" already used in our documentation: `/docs/monitoring-your-api/uptime-monitors/` and `/docs/getting-started/navigating-postman/`. Since we're already doing this in some places, it might be OK to continue doing this. Maybe we could leave it as "Paid plans only" for now, and then add this topic to a future style guide discussion?

anjel647: tw-821-workplace-roles-update

petewitcher: Personally I'd use the shorthand version (`Paid plans only`) to save space if necessary. 

I think `Paid plans only` works because I'm betting the user will know if they (or their employer) are paying for the plan. If they need more info about the plans, they have to click a link. A bit more work for them, but worth the risk IMO if I really want that extra space.

petewitcher (approved):
LGTM

src/pages/docs/collaborating-in-postman/roles-and-permissions.md
HeatherStenson: Thanks for the convo, all! In the interest of expediency, I'm going to leave this as "paid plans only" for now, and we can hash it out in the style guide backlog and standardize across the LC later.

PR #4482, opened 2022-12-22
Updated the link to FAQs directory

Raj-postman:
The older Postman app section is no longer present, which leads to the error screen "The page you are looking for no longer exists"


claireriley (approved):
Thanks for catching this @Raj-postman ! ⭐

PR #4484, opened 2022-12-23
Update 'adding import paths' section in 'using service definition'

skatikia:
'Adding import paths' section contains further clarification on how users can configure import paths based on the  type of import directive mentioned in the schema

Screen Shot 2022-12-23 at 4 33 40 PM


src/pages/docs/sending-requests/grpc/using-service-definition.md
```diff
@@ -80,7 +80,22 @@ Instead of importing a `.proto` file from your local system, you can also fetch
...
+```
+
+import "enums/NumericEnum.proto"
+import "messages/EmptyMessage.proto"
+import "messages/HelloResponse.proto"
+import "messages/HelloRequest.proto"
+
+```
+
+In this scenario, you'll need to configure the parent directory of `enums` and `messages` i.e. `proto` as your import path in Postman. The added import path can look something like this: `/Users/johndoe/projects/my-app/protos/`. You can configure import paths in a request using the **Import paths option** in the **Import flow** after you select the file to import.
```
appurva21, line 98: i.e. `protos`

appurva21 (approved):

claireriley (approved):
Thanks @skatikia ! ⭐ Made some very minor wording edits but LGTM. We'll include it in our next push to prod

PR #4507, opened 2023-01-10
User feedback

Raj-postman:
Including above details as suggested by user, feel free to have it edited or removed if it doesn't qualify to be present on the docs. Thank you!

> Just letting your users know that only Linux x86 is supported saves people a lot of time, frustration, and probably some emails and forum posts.


jkonrath-postman:
I've restarted this over here: https://github.com/postmanlabs/postman-docs/pull/4508 and as TW-840.

PR #4558, opened 2023-02-07
[TW-899] Doc OAuth support for ID tokens

mark-anthony-postman:
Added new section about changing token type (Access token or ID token)

### DIFF
--- a/src/pages/docs/sending-requests/authorization.md
+++ b/src/pages/docs/sending-requests/authorization.md
@@ -65,6 +65,7 @@ You can pass auth details along with any request you send in Postman. Auth data 
         * [Requesting an OAuth 2.0 token](#requesting-an-oauth-20-token)
         * [Refreshing an OAuth 2.0 token](#refreshing-an-oauth-20-token)
         * [Sharing an OAuth 2.0 token](#sharing-an-oauth-20-access-token)
+        * [Changing the OAuth 2.0 token type](#changing-the-oauth-20-token-type)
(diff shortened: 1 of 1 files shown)


HeatherStenson (approved):
Looks great! :fishsticks:

src/pages/docs/sending-requests/authorization.md
```diff
@@ -334,6 +335,14 @@ To revoke other users' access to a synced token, do the following:
 
 After you revoke access, other users with access to the request won't be able to see or use the token.
 
+#### Changing the OAuth 2.0 token type
+
+Postman supports using access tokens or ID tokens for OAuth 2.0 authorization. An _access token_ enables an OAuth client to make calls to an API. An _ID token_ contains information about the authenticated user. This information can be used by an OAuth client to customize their experience.
+
+To change the type of token used for authorization, select **Access token** or **ID token** in the **Use Token Type** dropdown list.
```
giridharvc7, line 342: Should we mention that the this dropdown option is available only if there's an id token present?

mark-anthony-postman: @giridharvc7 Yes, great suggestion! I've changed the wording here to:

If an ID token is present, you can select the token type (**Access token** or **ID token**) in the **Use Token Type** dropdown list. (If no ID token is present, this dropdown list isn't available.)

jkonrath-postman (approved):
LGTM

PR #4574, opened 2023-02-13
TW-939  Doc Create collection when starting proxy capture session

jkonrath-postman:


src/pages/docs/sending-requests/capturing-request-data/capturing-http-requests.md
```diff
@@ -98,7 +98,7 @@ To start a proxy debug session, do the following:
 
     > You can also capture cookies and sync them to Postman without starting a debug session. Learn more about [syncing cookies](/docs/sending-requests/capturing-request-data/syncing-cookies/).
 
-1. Requests will be saved in a debug session in the **History** tab in the sidebar. Select **Save requests to a collection** and select a collection from the list to also save the requests there.
+1. Requests will be saved in a debug session in the **History** tab in the sidebar. Select **Save requests to a collection** and select a collection from the list to also save the requests there. You can also type the name of a new collection and select the Create action to create a new collection.
```
HeatherStenson, line 101:
```suggestion
1. Requests will be saved in a debug session in the **History** tab in the sidebar. Select **Save requests to a collection** and select a collection from the list to also save the requests there. To create a new collection in which to save requests, enter a name for the new collection and select the Create action.
```

HeatherStenson (approved):
Looks good! One small suggestion on wording. (The only strongly held opinion in my suggestion is using "enter" instead of "type", per the style guide, everything else is up to you)

MatthewPinsker-postman (approved):
Looks good! As Heather mentioned, it looks like we prefer "enter" over "type" according to the A-Z word list.

PR #4593, opened 2023-02-17
[TW-946] Editing a gRPC example

petewitcher-postman:


src/pages/docs/sending-requests/grpc/using-grpc-examples.md
```diff
@@ -69,22 +69,21 @@ This walkthrough creates and executes a client streaming request, then saves the
...
-1. Select the message again to close the editor.
-
-1. Hover over a message. A handle appears to the left of the message. Select and drag the handle to reposition the message in the timeline.
-
-    ![Reposition the message](https://assets.postman.com/postman-docs/v10/grpc-save-example-4reposition-2-v10.jpg)
+    ![Status code 2 Unknown](https://assets.postman.com/postman-docs/v10/grpc-save-example-2unknown-1-v10.jpg)
 
 1. Select **Save**.
 
+This walkthrough and the one before it use unary methods. The next walkthrough gives you a look at a streaming method and how its messages differ from unary methods.
```
MatthewPinsker-postman, line 85: In the walkthrough before this one, it looks like we don't mention "unary methods". Would it be helpful to also mention in the previous walkthrough that it uses unary methods? I'm asking in case users might want to know that the previous one uses unary methods, but they don't read this walkthrough.

Also, what do you think about moving the second sentence to the next walkthrough, using it as an introduction? I'm asking in case users might find this information helpful in the context of the next walkthrough. Something like: `This walkthrough gives you a look at a streaming method and how its messages differ from unary methods, which are used in [Saving a gRPC examples](#saving-a-grpc-example) and [Editing a gRPC examples](#editing-a-grpc-example).`

MatthewPinsker-postman (approved):
Looks good to me!

src/pages/docs/sending-requests/grpc/using-grpc-examples.md
petewitcher-postman: Great suggestion, thanks. I had neglected to change `client streaming` to `unary` in the walkthrough before this one. Now the first walkthrough's intro states it uses a unary method. WRT the second suggestion, I think I accomplished the goal by editing and adding  intro sentences for the last two walkthroughs. Let me know if you think that works.

src/pages/docs/sending-requests/grpc/using-grpc-examples.md
```diff
@@ -65,28 +65,29 @@ This walkthrough creates and executes a client streaming request, then saves the
...
+    The response updates automatically.
 
-    ![Reposition the message](https://assets.postman.com/postman-docs/v10/grpc-save-example-4reposition-2-v10.jpg)
+    ![Status code 2 Unknown](https://assets.postman.com/postman-docs/v10/grpc-save-example-2unknown-1-v10.jpg)
 
 1. Select **Save**.
 
 ## Creating a gRPC example from scratch
 
+This walkthrough uses a streaming method to create a gRPC example. Note how streaming methods' messages differ from unary methods' messages.
```
MatthewPinsker-postman, line 89: This is minor, but the second sentence might be easier to scan if we remove the apostrophes. Something like: `Note how a streaming method message differs from a unary method message.`

src/pages/docs/sending-requests/grpc/using-grpc-examples.md
MatthewPinsker-postman: Looks good! I added one minor comment.

src/pages/docs/sending-requests/grpc/using-grpc-examples.md
petewitcher-postman: Agreed, done.

PR #4596, opened 2023-02-21
[MARKENG-2235] Add runtime check test for Navbar and Footer API

brandonmcastillo:
This update will check if the keys returned from the bff match the expected values and act as a safeguard if the data comes back malformed.

If true, the production data will be set and fallback to the local version if false.


ch264 (approved):
LGTM

PR #4606, opened 2023-02-26
update (some) of the redirects.json

tristandenyer:
* manually ran about 80 of the `from` URLs, let resolve, and updated the redirects where applicable to reduce the multi-redirect situations
* this is for the app-level redirects


jkonrath-postman (approved):
LGTM

PR #4613, opened 2023-02-28
move fonts to Voyager

tristandenyer:
* moved fonts to Voyager
* removed font files
* removed imports
* zero changes to front-end / UI should be present. Zero shift in fonts printing should be evident.


ch264 (approved):
pulled, ran, lgtm

PR #4664, opened 2023-03-20
[TW-972] Flows - Output block

petewitcher-postman:


jkonrath-postman:
Merged this into TW-977 so let's pick it up over in #4669

PR #4690, opened 2023-03-30
Fixing leftnav accordion hard load

Ryan-McBride:
Current implementation of the LeftNav in the /docs/ paths causes a hard browser navigation when an accordion is clicked to expand.

Current Implementation:
current

This MR modifies the CaretWrapper styled component to be a `Link` instead of an `a`, which does a proper in-app navigation.

It also changes the sectionHandler function to use Gatsby's built in `navigate` feature rather than the browser api for changing current url.

Improved Implementation:

improved


ch264 (approved):
lgtm 🦄

PR #4735, opened 2023-04-12
[TW-1061] No flows editor role in free plan

petewitcher-postman:


src/pages/docs/postman-flows/gs/flows-overview.md
```diff
@@ -61,7 +61,7 @@ These workspaces contains example flows utilizing a multitude of APIs for differ
 
 ## Enabling Postman Flows
 
-To get started using Postman Flows, you'll need to assign the Flow Editor role to one or more of your team members. For more information on adding a role to a user, see [Managing team roles](/docs/administration/managing-your-team/managing-your-team/#managing-team-roles).
+To get started using Postman Flows, you'll need to assign the Flow Editor role to one or more of your team members in a paid or early-access plan (The Flow Editor role is not available in the Postman Free plan). For more information on adding a role to a user, see [Managing team roles](/docs/administration/managing-your-team/managing-your-team/#managing-team-roles).
```
MatthewPinsker-postman, line 64: In the [Team roles section](https://learning.postman.com/docs/collaborating-in-postman/roles-and-permissions/#team-roles), we explain that the Flow Editor role is only available on Annual Professional plans, not all paid plans.

Should we update these two sections to match?

jkonrath-postman: No - the assumption is that if they are on a paid plan and they can use flows, this applies. (There might be edge cases when they might still be able to use it, not sure.)

jkonrath-postman, line 64:
```suggestion
To get started using Postman Flows in a paid or early-access plan, you'll need to assign the Flow Editor role to one or more of your team members. For more information on adding a role to a user, see [Managing team roles](/docs/administration/managing-your-team/managing-your-team/#managing-team-roles).
```

jkonrath-postman: We can simplify this and not specifically say this is a thing for free plans. It will just work for them, and no need to do this step.

jkonrath-postman (approved):
Minor change, but otherwise LGTM.

src/pages/docs/postman-flows/gs/flows-overview.md
```diff
@@ -61,7 +61,7 @@ These workspaces contains example flows utilizing a multitude of APIs for differ
 
 ## Enabling Postman Flows
 
-To get started using Postman Flows, you'll need to assign the Flow Editor role to one or more of your team members. For more information on adding a role to a user, see [Managing team roles](/docs/administration/managing-your-team/managing-your-team/#managing-team-roles).
+To get started using Postman Flows in a paid or early-access plan, you'll need to assign the Flow Editor role to one or more of your team members. For more information on adding a role to a user, see [Managing team roles](/docs/administration/managing-your-team/managing-your-team/#managing-team-roles).
```
cyrielle-hada, line 64: We can't really say in an "early-access plan", as early access is not the name of one of our Postman plans. Also from the paid plans for which Flows is available, Flows is only available to Professional plans so if we say "paid plans" it gives the impression Enterprise and Basic have access to Flows as well. 

Can we add a disclaimer as a note info box above the paragraph that specifies that the Flow editor role is only available to users who had access to Flows as early access and to Professional plans?

jkonrath-postman: Let's change this to "in a Professional plan" and leave it at that. This step is just a workaround step for teams who couldn't use Flows at all because they had no users with the role. We shouldn't get into who can and can't do this based on early access or other edge cases.

PR #4787, opened 2023-05-11
TW-1104 Account vs Team

claireriley:


jkonrath-postman (approved):
One little thing, but otherwise LGTM

src/pages/docs/collaborating-in-postman/working-with-your-team/collaboration-overview.md
```diff
@@ -31,13 +31,25 @@ To get the most out of Postman's collaboration features, [create](#creating-a-te
...
 
+* [About accounts and teams](#about-accounts-and-teams)
 * [Creating a team](#creating-a-team)
 * [Joining a team](#joining-a-team)
 * [Switching between teams](#switching-between-teams)
 * [Leaving a team](#leaving-a-team)
 * [Team usage limits](#team-usage-limits)
 * [Next steps](#next-steps)
 
+## About accounts and teams
```
MatthewPinsker-postman, line 42: Thanks for working on this!

I think we're using "Postman account", "personal account", and "individual account" interchangeably in a few places.

In the first paragraph, what do you think about adding something like: `When you sign up for Postman as an individual, you create a [Postman account](/docs/getting-started/postman-account/) and interact with Postman in an individual account, which only you can access. In your individual account, you can create requests, collections, APIs, and other Postman elements.` It's my understanding that an individual account is like a team but only you can access it, and my thinking is that we can clarify that here (if my understanding is correct).

In the last paragraph, should we be referring to an "individual account", instead of a "Postman account"? I might be thinking of something different, but I wanted to bring it up just in case.

claireriley: Hmm it's tricky to word it, I tried to use those terms for emphasis/clarity based on placement (or where I was linking out to other docs). Sometimes teams are worded as accounts and I want to make sure those concepts are kept separate for the user as I think it could confuse them. 

I think Arjun wanted to just call it "account", so I've removed/updated some of the descriptors in the upcoming commit to try to clean it up a bit.

MatthewPinsker-postman: OK! Sounds good!

MatthewPinsker-postman (approved):
Looks good!

src/pages/docs/collaborating-in-postman/working-with-your-team/collaboration-overview.md
```diff
@@ -31,13 +31,25 @@ To get the most out of Postman's collaboration features, [create](#creating-a-te
...
+
+When you sign up for Postman as an individual, you create a [Postman account](/docs/getting-started/postman-account/). If you [join a Postman team](#joining-a-team), you can decide to either move your workspaces and transfer ownership to the team or keep your workspaces separate and maintain ownership of them in your personal account. This decision affects your account:
+
+* When you move your workspaces to a team, your account will cease to exist and you'll only have the option to switch between teams, rather than work in a personal account.
+* When you keep your workspaces, you'll retain your account and you can switch between the team and your personal account at any time.
+
+In certain cases, your personal workspaces and the data within them automatically transfer when you join a team. For more information, see [Joining a team](#joining-a-team).
+
+When you [leave a team](#leaving-a-team), a [Postman account](/docs/getting-started/postman-account/) is created for you if you don't already have one. If you as a user are deactivated through SCIM, you'll need to [create a new Postman account](/docs/getting-started/postman-account/#signing-up-for-a-postman-account).
```
ArjunSingh-PM, line 51: We can remove this statement:

If you as a user are deactivated through SCIM, you'll need to [create a new Postman account](/docs/getting-started/postman-account/#signing-up-for-a-postman-account).

This is not entirely true. If you are deactivated in a team via SCIM, you will still have access to the personal account and other teams you are a member of. 

If the user is not a member of any other team and does not have a personal account either, their account will be locked-out of Postman i.e. they will not be able to authenticate into their account. They need to contact support in this case.

ArjunSingh-PM (approved):
The changes look good. Please review the comments before merging the change.

PR #4808, opened 2023-05-19
[TW-1153] Maintenance 5/23 updates

akinard-postman:
This updates the `/docs/integrations/available-integrations/azure-devops` documentation. Namely:
- Cleaning up the frontmatter.
- Added new frontmatter prereq link to the "Using a remote Git repository" doc.
- Changing the link for "Postman to Azure DevOps" integration to point at the on-prem section of the "Using a remote Git repository" doc. It seems like it makes more sense since the Azure DevOps document focuses on an on-premises integration.
- Updated the "API sync with Azure DevOps":
    - Removed the v10 specific mention — I'm not sure if that is necessary any longer?
    - Linked out to the API Builder at its mention.
    - Split the note into two paragraphs so it was a little easier to read.

### DIFF
--- a/src/pages/docs/integrations/available-integrations/azure-devops.md
+++ b/src/pages/docs/integrations/available-integrations/azure-devops.md
@@ -18,7 +18,7 @@ Postman enables you to sync your Postman APIs to Git repositories hosted in Micr
 
 ## API sync with Azure DevOps
 
-You can connect an Azure DevOps repository to an API in Postman's [API Builder](https://learning.postman.com/docs/designing-and-developing-your-api/the-api-workflow/). Once connected, you can sync your API's definition and associated collections between Postman and Azure DevOps. You can switch branches, pull changes from the repository, and push changes to the repository, all from within Postman. To learn more about syncing your API with Azure DevOps, see [Using a remote Git repository](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/).
+With Postman v10, you can connect an Azure DevOps repository to an API in Postman's [API Builder](https://learning.postman.com/docs/designing-and-developing-your-api/the-api-workflow/). Once connected, you can sync your API's definition and associated collections between Postman and Azure DevOps. You can switch branches, pull changes from the repository, and push changes to the repository, all from within Postman. To learn more about syncing your API with Azure DevOps, see [Using a remote Git repository](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/).
 
 > **To connect your API to Azure DevOps, make sure to enable third-party application access for your organization.** If you don't enable third-party access, Postman won't be able to connect to your repository.
 >
--- a/src/pages/docs/integrations/available-integrations/azure-devops.md
+++ b/src/pages/docs/integrations/available-integrations/azure-devops.md
@@ -1,12 +1,12 @@
 ---
 title: "Azure DevOps"
-order: 169
-page_id: "azure-devops"
-updated: 2021-12-20
-warning: false
+updated: 2023-05-19
 contextual_links:
   - type: section
     name: "Prerequisites"
+  - type: link
+    name: "Using a remote Git repository"
+    url: "/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/"
   - type: link
     name: "Grouping requests in collections"
     url: "/docs/sending-requests/intro-to-collections/"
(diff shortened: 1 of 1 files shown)


mark-anthony-postman (approved):
LGTM!

src/pages/docs/integrations/available-integrations/azure-devops.md
```diff
@@ -1,23 +1,25 @@
...
+    url: "/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/"
   - type: link
     name: "Grouping requests in collections"
     url: "/docs/sending-requests/intro-to-collections/"
 ---
 
 > **[Azure DevOps Server (hosted on-premises) integrations are available on Postman Enterprise plans.](https://www.postman.com/pricing)**
 
-Postman enables you to sync your Postman APIs to Git repositories hosted in Microsoft Azure DevOps with the [Postman to Azure DevOps integration](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/).
+Postman enables you to sync your Postman APIs to Git repositories hosted in Microsoft Azure DevOps with the [Postman to Azure DevOps integration](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/#connecting-to-an-on-premises-repository).
```
MatthewPinsker-postman, line 17: Should we keep this link pointing to the top of the file, instead of the on-premises sub-section? I'm asking because it looks like users can create an integration either hosted on-prem or in the cloud.

src/pages/docs/integrations/available-integrations/azure-devops.md
```diff
@@ -1,23 +1,25 @@
...
 
 ## API sync with Azure DevOps
 
-With Postman v10, you can connect an Azure DevOps repository to an API in the API Builder. Once connected, you can sync your API's definition and associated collections between Postman and Azure DevOps. You can switch branches, pull changes from the repository, and push changes to the repository, all from within Postman. To learn more about syncing your API with Azure DevOps, see [Using a remote Git repository](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/).
+You can connect an Azure DevOps repository to an API in Postman's [API Builder](https://learning.postman.com/docs/designing-and-developing-your-api/the-api-workflow/). Once connected, you can sync your API's definition and associated collections between Postman and Azure DevOps. You can switch branches, pull changes from the repository, and push changes to the repository, all from within Postman. To learn more about syncing your API with Azure DevOps, see [Using a remote Git repository](/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/).
```
MatthewPinsker-postman, line 21: Do we want to move away from the format "With Postman v10, you can ..."? I'm asking because we have similar sections in the GitHub, GitLab, and Bitbucket integration pages that still use this format. Not sure if we intended to follow this format to differentiate features between different versions, or if we want to make this update in all 4 pages in one sweep.

akinard-postman: @MatthewPinsker-postman Yeah I was a bit uncertain about that. What I can do here is revert this back for now. I'll open a ticket around this for discussion. Thank you for the insight here!

MatthewPinsker-postman: Sounds good! Thanks!

src/pages/docs/integrations/available-integrations/azure-devops.md
akinard-postman: @MatthewPinsker-postman I changed this to point at this specific section based around the note at the top of the document that says that "Azure DevOps Server (hosted on-premises) integrations...". 

Should we maybe point this link to https://learning.postman.com/docs/designing-and-developing-your-api/versioning-an-api/using-external-git-repo/#connecting-to-azure-devops-server instead?

MatthewPinsker-postman: I might be mistaken, but I think the note was added at the top because an Enterprise account is required for on-prem, and there aren't plan restrictions for cloud. The new link you added would be more specific to the source control system, but I think it's still specific to on-prem.

One thing to mention is that we have a second link to the same page (/using-external-git-repo) after the H2. It might be helpful to users if these links either point to the same section on the page, or we only have one link to the page.

akinard-postman: I think your suggestion of having just one link to the page would be a better path here. I'll go with that suggestion. Thank you so much for the help!

MatthewPinsker-postman: No problem!

PR #4816, opened 2023-05-22
Add overflow css to <pre> tag

mark-anthony-postman:
Fixes issue where long lines overflow to the right in code reference tables when using the `<pre>` tag.

image

Solution is to add the following css to the `<pre>` tag in `/src/templates/doc.jsx`:

```
overflow-x: auto;
white-space: pre-wrap;
word-wrap: break-word;
```

Tested the fix in Chrome, Safari, and Firefox.

Source of this solution: https://www.w3docs.com/snippets/css/how-to-wrap-text-in-a-pre-tag-with-css.html

### DIFF
--- a/src/templates/doc.jsx
+++ b/src/templates/doc.jsx
@@ -111,6 +111,9 @@ pre {
   font-family: 'IBM Plex Mono',SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace !important;
   line-height: 24px;
   margin-bottom: 0px;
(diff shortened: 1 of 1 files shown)


petewitcher-postman (approved):
LGTM!

DanielKimmelmann1 (approved):
👍

ch264 (approved):
pulled ran, also tested on Edge, looks good to me

PR #4855, opened 2023-06-06
[markeng-2488] add catch to Agolia for index switching

ch264:
This PR adds a catch to display a message to the user if a index endpoint cannot be found and prevent the app from breaking.

Screenshot 2023-06-06 at 3 56 25 PM


jkonrath-postman (approved):
LGTM

PR #4871, opened 2023-06-09
Improve pageload performance with better asset loading

tristandenyer:
* saw in page speed test that the fonts were render blocking
* using this SO answer: https://stackoverflow.com/questions/45983301/google-pagespeed-eliminate-render-blocking-resources-above-the-fold-caused-fro#answers-header
* Improved font asset loading
* and removed the redirect polyfil was performing.
* this improves the performance of the page load speeds for all of LC by an estimated 1sec


src/components/seo.jsx
```diff
@@ -112,7 +113,7 @@
        {/* Algolia Instantsearch IE11 support v4 */}
        <link rel="dns-prefetch" href="https://polyfill.io" />
        <link crossOrigin rel="preconnect" href="https://polyfill.io" />
-       <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.find%2CArray.prototype.includes%2CPromise%2CObject.assign%2CObject.entries"></script>
+       <script async src="https://polyfill.io/v3/polyfill.js?features=default%2CArray.prototype.find%2CArray.prototype.includes%2CPromise%2CObject.assign%2CObject.entries"></script>
```
oswish, line 116: Is it possible to add the async attribute but also load the .min version?

tristandenyer: @oswish We were getting yelled at about the `.min` version in that it was redirecting / failing.
Screenshot 2023-06-12 at 08 43 47

oswish: I found this:
https://polyfill.io/v3/supported-browsers
image

oswish (approved):
LGTM!

PR #4872, opened 2023-06-09
add async to OneTrust script to keep from blocking rendering

tristandenyer:
* added `async` to the OneTrust script
* suggested by page speed insights tool
* OT banner should still render as it has been—no changes
Screenshot 2023-06-09 at 12 19 16


ch264 (approved):
lgtm

PR #4955, opened 2023-07-24
[TW-1266] Audit plan/pricing info links

nerminamiller-postman:
Adds the enterprise flag to the API governance and security intro

### DIFF
--- a/src/pages/docs/api-governance/api-governance-overview.md
+++ b/src/pages/docs/api-governance/api-governance-overview.md
@@ -28,7 +28,7 @@ contextual_links:
     url: "https://blog.postman.com/5-ways-to-reduce-exposure-to-api-security-risks/"
 ---
 
-> **[API governance and API security are available on Postman Professional and Enterprise plans.](https://www.postman.com/pricing)**
+> **[API Governance and API Security are available on Postman Enterprise plans.](https://www.postman.com/pricing)**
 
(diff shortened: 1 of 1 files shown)


src/pages/docs/api-governance/api-governance-overview.md
```diff
@@ -28,6 +28,8 @@ contextual_links:
     url: "https://blog.postman.com/5-ways-to-reduce-exposure-to-api-security-risks/"
 ---
 
+> **[API governance and API security are available on Postman Professional and Enterprise plans.](https://www.postman.com/pricing)**
```
MatthewPinsker-postman, line 31: Thanks for working on this! I have a few suggestions and questions:

I think this should be changed to "API Governance and API Security". When we talk about the features, we use title case, such as "API Governance". (As I understand it) When we talk about governance in the context of APIs, we would use "API governance". Please let me know if you have any questions about this!

Are these features available on Professional plans? I just want to double check because I previously thought they were only available on Enterprise plans.

The pages in the `API definition` and `Requests` folders don't have pricing notes at the top of each page. Should we add a pricing note to some of these pages.

nerminamiller-postman: They are Enterprise, you are right! Will update that and the capitalization. Good feedback! Not sure about those pages. Maybe we should chat on Slack?

MatthewPinsker-postman: Thanks for double checking! Sure let's chat via Slack.

jkonrath-postman (approved):
LGTM

PR #4959, opened 2023-07-25
Update using-collections.md

Raj-postman:
Change in the recovery time frame for various plans, check Platform Security and Administration > Collection recovery in the page: https://www.postman.com/pricing


Raj-postman:
@jkonrath-postman It looks like the page had the wrong information due to last pricing update. Flagged it in the Slack channel earlier today and the team confirmed to update them soon.

So the original information is right, 30 days for basic and 90 days for Professional. Sorry for the false alarm!

PR #4982, opened 2023-08-02
fetch performance improvements on font css file and OT

tristandenyer:
- add back in the performance improvements we reverted when we got CORS issues.
- @mclausaudio fixed up CORS, so we should not see red on BETA for fetching the .css file
- Should no longer see this in Inspect > Performance test tab

Screenshot 2023-08-02 at 16 30 24


mclausaudio (approved):
LGTM

PR #5000, opened 2023-08-09
[TW-1234] Workspace template fixes

MatthewPinsker-postman:
In the initial release (July), we mentioned that users can select a workspace template (during workspace creation) only for Free, Basic, and Professional plans. This feature will be applied to Enterprise plans in the next release. For features like this, we typically don't include a pricing note, so we're removing the pricing notes in this PR.

* Removed pricing note from mentions about selecting a workspace template during workspace creation
* Updated the workspace creation CTA to "Create", which is the label used in this new workspace template flow
* Removed an extra note about adding a workspace template to a new workspace, which was redundant information


mark-anthony-postman (approved):
LGTM! 🌟

PR #5129, opened 2023-10-09
[MARKENG-2860] Deploy page-data.json files into Marketing website bucket

mclausaudio:
https://postmanlabs.atlassian.net/browse/MARKENG-2860

We need to deploy LC's page-data JSON (which contains the docs content as HTML) into WWW bucket, to be fetched across teams internally.


brandonmcastillo (approved):
LGTM

PR #5135, opened 2023-10-10
Markeng-2855  updated LC top navbar

ch264:
update top navbar


brandonmcastillo:
I found a few issues on the Secondary Nav.
- Text should be "Learning Center" 
- Algolia search was removed

brandonmcastillo (approved):
Pulled and ran locally. Compared it to Blog and LGTM

PR #5191, opened 2023-10-31
Move codex file into static folder, so it gets deloyed

mclausaudio:
Right now the Codex file is not getting deployed because it is in the root of the repo.  Since it is a static asset, it needs to go into `static` folder.  All assets in the `static` folder get deployed to the root of the `public` folder.


tristandenyer (approved):

tristandenyer:
@jkonrath-postman FYI, codex moved

PR #5197, opened 2023-11-03
reinstate missing package-lock; update GA npm

tristandenyer:
* the package-lock file was missing
* rebuilt it with latest
* and then ran `npm install gatsby-plugin-google-analytics@5.12.0` to update package to remove 1 vuln


tristandenyer:
will remake with latest

PR #5243, opened 2023-11-24
Docs: Convert It To Postman Collection Using Akash Amat's JS Tool

Mane-Olawale:


jkonrath-postman:
Hi @Mane-Olawale - thanks for the contribution, but for this workflow, we'd prefer users to use HAR import in Postman for this.

PR #5254, opened 2023-11-30
[TW-1490] Add note about the Workspace roles API

akinard-postman:
This adds a small callout on `collaborating-in-postman/user-groups.md` and `collaborating-in-postman/roles-and-permissions.md` that mention the Postman API.

This actually links to the main landing page of the Postman API in the Postman Public Workspace because the API doesn't exist yet. I can provide a subsequent update at a later date if needed that points directly to the specific collection folder (it will be Workspaces -> Roles).

**Note:**

I wasn't sure about updating the `updated` field in the frontmatter because it was such a minor update that I did not want to push the pages down the line for a refresh update.

### DIFF
--- a/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
+++ b/src/pages/docs/collaborating-in-postman/roles-and-permissions.md
@@ -1,7 +1,7 @@
 ---
 title: "Define roles and permissions within a Postman team"
 order: 72
-updated: 2022-12-19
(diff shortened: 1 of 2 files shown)


mark-anthony-postman (approved):
LGTM! And I agree, don't change the updated date. Save that for when you've gone through the whole page

jkonrath-postman (approved):
LGTM

PR #5269, opened 2023-12-05
[markeng2974] add a hidden page with list of early_access pages to th…

ch264:
added a hidden page with list of pages that have the enum (early_access:true)  to the docs folder


jkonrath-postman:
One little thing - the bulleted list has an entry at the end which is a link to the same page. Is there any way to tweak your query to make it not pick up that page?

image

jkonrath-postman (approved):
LGTM

PR #5303, opened 2023-12-18
Updating team-ids to TEAM_IDS on the MSI Installation Doc

Hollywood:
The standard behavior with .msi installers is that all-caps properties are considered Public and available in the command-line installations we have documented.

[Public Properties - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/msi/public-properties)

Properties that are lowercase are considered private and would not be available in the command-line install.

Also, fixing the spelling, as the currently documented property should have an underscore as opposed to a hyphen.

image


jkonrath-postman (approved):
LGTM. I'll merge.

PR #5427, opened 2024-02-27
[TW-1643] Unified comments

petewitcher-postman:
Updates inline comment functionality and key images. Working with Jon and Matthew to coordinate updating several other images with this and adjacent UI changes.


MatthewPinsker-postman (approved):
LGTM!

grakhejapstmn:
Let's omit the bit about pre requests and tests scripts. We are building a new interaction for that section alone which would be available in the post con release.
Here's a loom for what it will mostly look like https://www.loom.com/share/c4633057497749cc9b26f19c2125e4de

Also these changes would be disabled till post con so let's not make the support center changes before that.

